PR #4, opened 2019-08-01
[MOBILE-583] add custom event

crow:
### What do these changes do?

Adds custom event to flutter implementation.

To allow flutter developers to create and track custom events

Set up a custom event to generate and add itself on the enable puch checkbox action in main.dart:

```
var customEvent = new CustomEvent("hello", 1);
customEvent.interactionType = "whatever_type";
customEvent.interactionId = "who_am_i";
customEvent.setBoolProperty(true, "boolVal");
customEvent.setStringArrayProperty(["cool", "story"], "stringArrVal");
customEvent.setStringProperty("hello", "stringVal");
customEvent.setNumberProperty(1, "numVal");
//customEvent.properties = {"numVal":1, "stringVal":"hello", "stringArrVal":["cool", "story", 1], "boolVal":true };
Airship.addEvent(customEvent);
```

Then used debugger to ensure (on both iOS and Android) that all the fields were in the added final tracked event and any bad values were dropped or rendered the event invalid.


lib/src/custom_event.dart
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * Sets a custom String property. The value's length must not exceed 255 characters
+   * or it will invalidate the event.
+   *
+   * @param value The property value.
+   * @param key The property key.
+   */
+  void setStringProperty(String value, String key) {
```
rlepinski: Pretty sure you its standard practice to have the key as the first param

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -131,6 +133,73 @@ class AirshipPlugin : MethodCallHandler {
...
 
+    private fun addEvent(call: MethodCall, result: Result) {
+        val eventMap = call.arguments as HashMap<*, *>
+
+        val eventName = eventMap[CustomEvent.EVENT_NAME] as String? ?: run {
+            result.success(false)
+            return
+        }
+
+        CustomEvent.Builder(eventName).let {
```
rlepinski: You want to rewrite this a bit:

```
 val event = CustomEvent.Builder(eventName).let {
            val eventVal = (eventMap[CustomEvent.EVENT_VALUE] as Int?)
            if (eventVal != null) {
                it.setEventValue(eventVal)
            }

            val properties = eventMap[CustomEvent.PROPERTIES] as HashMap<String, Any>?
            if (properties != null) {
                parseProperties(properties, it)
            }

            val transactionId = eventMap[CustomEvent.TRANSACTION_ID] as String?
            if (transactionId != null) {
                it.setTransactionId(transactionId)
            }

            val interactionId = eventMap[CustomEvent.INTERACTION_ID] as String?
            val interactionType = eventMap[CustomEvent.INTERACTION_TYPE] as String?

            if (interactionId != null && interactionType != null) {
                it.setInteraction(interactionType, interactionId)
            }
        }
        .build()

if (event.isValid) {
                event.track()
                result.success(true)
            } else {
                result.success(false)
            }
```

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -196,3 +248,29 @@ UADeepLinkDelegate, UAPushNotificationDelegate, UAInboxDelegate {
         }
     }
 }
+
+private extension UACustomEvent {
+    func parseProperties(properties:Dictionary<String, Any>) {
```
rlepinski: You can do extensions on kotlin as well if you want the code to look the same

lib/src/custom_event.dart
```diff
@@ -0,0 +1,77 @@
...
+  void setBoolProperty(String key, bool value) {
+    _properties[key] = value;
+  }
+
+  /**
+   * Sets a custom String property. The value's length must not exceed 255 characters
+   * or it will invalidate the event.
+   *
+   * @param value The property value.
+   * @param key The property key.
```
rlepinski: need to update the doc comments

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -131,6 +133,69 @@ class AirshipPlugin : MethodCallHandler {
...
+            return
+        }
+
+        val event = CustomEvent.Builder(eventName).apply {
+            (eventMap[CustomEvent.EVENT_VALUE] as Int?)?.let {
+                this.setEventValue(it)
+            }
+
+            (eventMap[CustomEvent.PROPERTIES] as HashMap<String, Any>?)?.let {
+                parseProperties(it, this)
```
BrianBatchelder: I don't know Kotlin, but it seems like you could just use `this` inside `parseProperties()`, rather than passing it in?

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -131,6 +133,69 @@ class AirshipPlugin : MethodCallHandler {
...
+
+            if (value is Boolean) {
+                builder.addProperty(key, value)
+                continue
+            }
+
+            if (value is String) {
+                builder.addProperty(key, value)
+                continue
+            }
```
BrianBatchelder: Might be a coding style thing, but can these three `if`s be compressed into one with three `OR`s?

ios/Classes/SwiftAirshipPlugin.swift
crow: added

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
crow: not able to do that in the described situation as the function's 'this' would not be the same as 'this' inside the let block, updated this to an extension though which cleans it up

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
crow: explained in person

PR #8, opened 2019-09-20
[MOBILE-584] Add notification management

oristanovic:
### What do these changes do?
Adds notification management API to flutter implementation 

Allow developers to manage active notifications 

Added method calls to airship_bloc.dart 
`Airship.activeNotifications.then((val) {
      debugPrint('Active notifications: $val');
    });` 
then sent a push notification and used the debugger on iOS and Android to check if the API works as expected

I was trying to port the React Native implementation but am not too sure about the activeNotifications Android implementation and what the payload should contain

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -13,6 +13,7 @@ import com.urbanairship.util.DateUtils
 import com.urbanairship.util.UAStringUtil
 import com.urbanairship.widget.UAWebView
 import com.urbanairship.widget.UAWebViewClient
+import java.lang.NumberFormatException
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
@@ -21,8 +22,6 @@ import io.flutter.plugin.common.PluginRegistry.Registrar
 import io.flutter.plugin.common.StandardMessageCodec
 import io.flutter.plugin.platform.PlatformView
 import io.flutter.plugin.platform.PlatformViewFactory
-import java.lang.NumberFormatException
-
 
 class InboxMessageViewFactory(private val registrar: Registrar) : PlatformViewFactory(StandardMessageCodec.INSTANCE) {
     override fun create(context: Context, viewId: Int, arguments: Any?): PlatformView {
@@ -214,11 +213,10 @@ class AirshipPlugin : MethodCallHandler {
                 val statusBarNotifications = notificationManager.activeNotifications
 
                 for (statusBarNotification in statusBarNotifications) {
-
-                    val notification = HashMap<String, Any>()
-                    notification.put("notification_id", statusBarNotification.id.toString())
-
-                    notifications.add(notification)
+                    var message = Utils.shared.messageFromNotification(statusBarNotification)
+                    val tag = statusBarNotification.tag ?: ""
+                    val id = statusBarNotification.id.toString()
+                    notifications.add(Utils.shared.notificationObject(message, tag, id))
                 }
                 result.success(notifications)
             } else {
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -1,12 +1,16 @@
 package com.airship.flutter
 
+import android.app.NotificationManager
 import android.content.Context
 import android.view.View
+import android.os.Build
+import androidx.core.app.NotificationManagerCompat
 import com.urbanairship.UAirship
 import com.urbanairship.analytics.CustomEvent
 import com.urbanairship.json.JsonMap
 import com.urbanairship.json.JsonValue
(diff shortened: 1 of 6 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -198,6 +206,64 @@ class AirshipPlugin : MethodCallHandler {
...
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                val notifications = arrayListOf<Map<String, Any>>()
+
+                val notificationManager = UAirship.getApplicationContext().getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+                val statusBarNotifications = notificationManager.activeNotifications
+
+                for (statusBarNotification in statusBarNotifications) {
+
+                    val notification = HashMap<String, Any>()
+                    notification.put("notification_id", statusBarNotification.id.toString())
```
rlepinski: Looks like you are missing putting the original push body into the notification. Also seems broke on React.

This is how we accomplish it on Cordova - https://github.com/urbanairship/urbanairship-cordova/blob/master/src/android/CordovaNotificationProvider.java#L40
https://github.com/urbanairship/urbanairship-cordova/blob/master/src/android/UAirshipPlugin.java#L1045
https://github.com/urbanairship/urbanairship-cordova/blob/master/src/android/Utils.java#L31

oristanovic: Right, so if follow the same logic and set the flutter notification provider like this:
https://github.com/urbanairship/urbanairship-cordova/blob/639efa3ee0945cbdce80b06aa52a64af9e8051e4/src/android/CordovaAutopilot.java#L66
then since it takes the config as parameter, does that mean that something similar to this should be added to flutter plugin?
https://github.com/urbanairship/urbanairship-cordova/blob/639efa3ee0945cbdce80b06aa52a64af9e8051e4/src/android/PluginManager.java#L307

rlepinski: Nah, I think you can get away with just putting the payload in the extras for now. In FlutterAutopilot:
```
        airship.pushManager.notificationProvider = object : AirshipNotificationProvider(UAirship.getApplicationContext(), airship.airshipConfigOptions) {
            override fun onExtendBuilder(context: Context, builder: NotificationCompat.Builder, arguments: NotificationArguments): NotificationCompat.Builder {
                builder.extras.putBundle(PUSH_MESSAGE_BUNDLE_EXTRA, arguments.message.pushBundle)
                return super.onExtendBuilder(context, builder, arguments)
            }
        }
```

rlepinski:
This looks great. I think you need to rebase to get the CI to pass. Could you provide the output for both platforms so we can see where we are being inconsistent? Obviously the push payloads are going to be different, but we want the top level values to be consistent.

android/src/main/kotlin/com/airship/flutter/Utils.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.json.JSONException
+
+class Utils {
+
+    companion object {
+        val shared = Utils()
+    }
+
+    @RequiresApi(Build.VERSION_CODES.KITKAT)
+    fun messageFromNotification(statusBarNotification: StatusBarNotification): PushMessage {
```
rlepinski: You should either move these into the companion object to be static, or probably better would be to make an extension for the StatusBarNotfiication to pull out the PushMessage.

android/src/main/kotlin/com/airship/flutter/Utils.kt
```diff
@@ -0,0 +1,76 @@
...
+                continue
+            }
+            val value = message.pushBundle.getString(key)
+            if (value != null) {
+                extras[key] = value
+            }
+        }
+
+        if (message.alert != null) {
+            notification["alert"] = message.alert!!
```
rlepinski: you should not have to `!!` the value after doing a check above

android/src/main/kotlin/com/airship/flutter/Extensions.kt
```diff
@@ -47,4 +52,18 @@ fun CustomEvent.Builder.parseProperties(map:HashMap<String, Any>) {
             continue
         }
     }
+}
+
+@RequiresApi(Build.VERSION_CODES.KITKAT)
+fun StatusBarNotification.pushMessage(): PushMessage {
+
```
BrianBatchelder:
```suggestion
```

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -198,6 +205,63 @@ class AirshipPlugin : MethodCallHandler {
...
+                result.success(notifications)
+            } else {
+                result.error("UNSUPPORTED", "Getting active notifications is only supported on Marshmallow and newer devices.", null)
+            }
+
+    private fun clearNotification(call: MethodCall, result: Result) {
+        val identifier = call.arguments as String
+
+        if (UAStringUtil.isEmpty(identifier)) {
+            return;
```
BrianBatchelder: There needs to be a result call before each `return` statement.

PR #11, opened 2019-09-27
Fix events on Android

rlepinski:
The wrong autopilot was being used, so event handling was not working. I also noticed an error about notifying on a bg thread, so i fixed that by moving notify to the main thread with Kotlin coroutines.


android/src/main/kotlin/com/airship/flutter/EventManager.kt
```diff
@@ -27,6 +30,9 @@ class EventManager {
 
     fun notifyEvent(event: Event) {
         streams[event.eventType]?.notifyEvent(event)
```
crow: remove this line?

PR #12, opened 2019-09-29
[MOBILE-834] set up flutter ci

crow:
iOS:
Ran suggested xcode project update suggestions, did some general cleanup

Android:
Added a couple of suggested changes for getting the current example to build using flutter build command.

Specifically added some android X compatibility features to the gradle.properties:
```
android.useAndroidX=true
android.enableJetifier=true 
```

### What do these changes do?
Prepares the flutter project for running in ci setup.

Included are changes necessary to get the two primary flutter-based build commands working:

```
flutter build apk --release
flutter build ios --release --no-codesign
```

These build commands are what average users will likely use. They are outlined [here](https://flutter.dev/docs/deployment/cd)

Ran the job.

image


rlepinski:
The CI script part should be in the repo:
```
#!/usr/bin/env bash
# fail if any commands fails
set -e
# debug log
set -x

cd example

echo "Building iOS Sample"
flutter build ios --release --no-codesign

echo "Building Android Sample"
flutter build apk --release
```

example/android/scripts/run_ci_tasks.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
rlepinski: do we need this file

scripts/run_ci_tasks.sh
```diff
@@ -0,0 +1,83 @@
...
+ANDROID=true
+ANALYZE=true
+IOS=true
+
+# Parse arguments
+OPTS=`getopt haiz $*`
+if [ $? != 0 ] ; then echo "Failed parsing options." >&2 ; exit 1 ; fi
+eval set -- "$OPTS"
+
+while true; do
```
rlepinski: Can we remove this optional command for now?

scripts/run_ci_tasks.sh
```diff
@@ -0,0 +1,83 @@
...
+done
+
+# verify flutter is installed and correctly configured
+flutter --version
+flutter doctor
+flutter packages get
+
+# set up flutter project
+if [[ "$BITRISE_SOURCE_DIR" != "" ]]; then
+    REPO_PATH="${BITRISE_SOURCE_DIR}"
```
rlepinski: Do we really need this? I would rather not have bitrise stuff in the script

crow: yeah, probably unnecessary

example/android/scripts/run_ci_tasks.sh
crow: nope

scripts/run_ci_tasks.sh
crow: going to keep it because it makes it easier to troubleshoot locally

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -366,7 +364,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
```
BrianBatchelder: If this is targeting SDK 12, the target should be 11.

crow: we aren't targeting SDK 12 yet

PR #15, opened 2019-10-01
[MOBILE-849] Flutter sample - Deep linking, message views

crow:
### What do these changes do?

- Restructures the sample to use Futures directly from airship instead of consolidating all the airship functionality into a BLoC class and using streams.  This allows functionality that needs to be distributed across views simpler to implement.

- Adds deep linking functionality

- Fixes a message center bug causing messages to "return" after being deleted (they weren't actually being properly removed from airship)

To make the sample nicer and to illustrate basic deep linking in the flutter sample.

Manually tested on iOS and Android (see video below).

https://slack-files.com/T025Q1VP7-FNZ2DCH7G-eac954900e


android/src/main/kotlin/com/airship/flutter/EventManager.kt
```diff
@@ -45,7 +46,7 @@ internal class AirshipEventStream(eventType: EventType) : EventChannel.StreamHan
     fun notifyEvent(event: Event) {
         var sink = eventSink
         if (sink != null) {
-            sink.success(event.eventBody?.toString())
+            sink.success(event.eventBody?.getString(event.eventBody.toString()))
```
BrianBatchelder: That looks really weird. What is happening here?

example/lib/screens/home.dart
```diff
@@ -1,10 +1,28 @@
...
+  void initState() {
+    initAirshipListeners();
+    super.initState();
+  }
+
+  // Platform messages are asynchronous, so we initialize in an async method.
+  Future<void> initAirshipListeners() async {
+    Airship.onChannelRegistration.listen((event) {
+      if (mounted) {
+        setState(() {});
```
BrianBatchelder: Can you also explain this one? Thanks for teaching me Dart.

example/lib/screens/message_center.dart
```diff
@@ -23,17 +40,18 @@ class MessageCenter extends StatelessWidget {
...
-              _airshipBloc.messageRemovedSink.add(message);
+              Airship.deleteInboxMessage(message);
+              setState(() {});
             },
             // Add stream to check isRead
             child: ListTile(
               title: message.isRead ? Text('${message.title}') : Text('${message.title}', style:TextStyle(fontWeight: FontWeight.bold)),
               subtitle: Text('${message.sentDate}'),
-              leading: Icon(message.isRead ? Icons.markunread : Icons.check_circle) ,
+              leading: Icon(message.isRead ? Icons.check_circle: Icons.markunread),
```
BrianBatchelder: Is this right? It looks like you will display `check_circle` when the message `isRead`, and `markunread` when the message is not `isRead`, which I'm interpreting as `isUnread`. It seems like you'd want it the way it was before.

example/lib/screens/settings.dart
```diff
@@ -18,20 +21,19 @@ class Settings extends StatelessWidget {
...
                         style: Styles.settingsPrimaryText,),
-                      value: snapshot.hasData ? snapshot.data : false,
-                      onChanged: (bool value){
-                        _airshipBloc.notificationsEnabledSetSink.add(value);
-                        },
-                    );
-                    }),
-                StreamBuilder(
-                    stream: _airshipBloc.namedUserStream,
+                      value: snapshot.data ?? false,
```
BrianBatchelder: Should this have changed to `snapshot.data`? I think that can be a different type than boolean, but I'm not sure what effect that would have.

android/src/main/kotlin/com/airship/flutter/EventManager.kt
crow: getString is a json parsing method that requires a fallback string. The previous implementation was just passing toString which isn't properly parsed and comes out as ""contents"" instead of "contents". I wasn't sure what to put as the default in this case. We could also put an empty string or something. I figured we'd fall back to the guaranteed toString output in the event a non-json encoded string body is provided.

example/lib/screens/home.dart
crow: You can infer if the widget is still displaying by checking its mounted flag. In this case, we avoid trying to set the state of the widget when its unmounted (aka. no longer visible) since we can receive this async Airship callbacks at odd times in the widget's lifecycle.

example/lib/main.dart
```diff
@@ -30,14 +35,58 @@ class _MyAppState extends State<MyApp> with SingleTickerProviderStateMixin {
...
+        .listen((event) => debugPrint('Inbox updated link'));
+
+    Airship.onShowInbox
+        .listen((event) => debugPrint('Show inbox'));
+
+    Airship.onShowInboxMessage
+        .listen((messageId) => debugPrint('Show inbox message $messageId'));
+
+    Airship.onChannelRegistration.listen((event) {
+      debugPrint('Channel registration $event');
```
marc-scig: minor formatting nit: do we want indentation to be two or four spaces?

android/src/main/kotlin/com/airship/flutter/EventManager.kt
BrianBatchelder: OK.

example/lib/screens/home.dart
BrianBatchelder: What does `() {}` do?

example/lib/screens/message_center.dart
marc-scig: If markunread doesn't imply that the message is unread, what does it imply?

crow: This behavior is correct. Check circle appears when the message isRead, markunread (an unopened envelope icon) appears otherwise. Somehow I had this swapped last time. You can confirm this functionality in the video.

marc-scig: :cool:

example/lib/screens/settings.dart
crow: this future can only result in a null or a boolean, if it's null is coalesces to false

android/src/main/kotlin/com/airship/flutter/EventManager.kt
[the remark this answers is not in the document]
crow: explained above

example/lib/screens/home.dart
crow: setState notifies the UI that the internal state of the widget has changed. It has a block that can be supplied to key off specific properties on the widget, in this case we want to do a complete rebuild so the setter block is left blank.

Some dart people consider this a code smell. I think this is true for projects that need to key state changes carefully to prevent unnecessary refreshes; but I disagree that it's a problem in this instance because the potential performance hit is negligible to the point of being irrelevant in our use case.

android/src/main/kotlin/com/airship/flutter/EventManager.kt
rlepinski: I think you could just do optString right?

example/lib/main.dart
crow: I copied and re-pasted this and this is what it auto-formats to. The auto formatting indents the .listen call over slightly farther. Since this is what it does automatically I'm not going to mess with it.

PR #16, opened 2019-10-03
Mobile 587 add in app automation support

BMatthai:
### What do these changes do?
It exposes paused property for In-App automation through getter and setter.

They are necessary to support In-App automation.

I didn't verify changes. I don't know how to do it on Flutter.

None

None


rlepinski:
This looks great. The only thing we might want to change is the naming. @BrianBatchelder have opinions on the method names?

BrianBatchelder:
I think `setInAppAutomationPaused` and `getInAppAutomationPaused` would better match the existing `setUserNotificationsEnabled` and `getUserNotificationsEnabled` methods.

PR #17, opened 2019-10-07
[MOBILE-585] Add tag group support

oristanovic:
### What do these changes do?
Adds tag group API to flutter implementation

Allow developers to manage tag groups

Added the following code to main.dart

```
    List<TagGroupOperation> operations = new List();

    AddTagGroupOperation operation = new AddTagGroupOperation("test_group", ["tag4", "tag5"]);
    operations.add(operation);

    Airship.editChannelTagGroups(operations).then((val) {
      debugPrint('operations: $operations');
    });
```
then used the debugger on iOS and Android to check if the API works as expected and checked that the new tags are available for segmentation


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -180,6 +189,49 @@ class AirshipPlugin : MethodCallHandler {
...
+
+            var group = operation[TAG_OPERATION_GROUP_NAME] as String ?: continue
+            var tags = operation[TAG_OPERATION_TAGS] as ArrayList<String?> ?: continue
+            var operationType = operation[TAG_OPERATION_TYPE] as String ?: continue
+
+            var tagSet = mutableSetOf<String?>()
+            for (j in 0 until tags.size) {
+                var tag = tags[j] as String?
+                if (tag != null) {
+                    tagSet.add(tags[j])
```
BrianBatchelder:
```suggestion
                    tagSet.add(tag)
```

lib/src/airship.dart
```diff
@@ -181,6 +181,20 @@ class Airship {
     return await _channel.invokeMethod('removeTags', tags);
   }
 
+  static Future<void> editChannelTagGroups(List<Map<String, dynamic>> operations) async {
```
rlepinski: Can we make a model object to define the operations sorta like we do in react native?

oristanovic: From what I can see, the operations are defined like this in react native
`var operation = { "operationType": "add", "group": group, "tags": tags }`
You mean create something like a tag group editor? https://github.com/urbanairship/react-native-module/blob/master/js/TagGroupEditor.js

rlepinski: Yes, the tag group editor. That way the underlying operation payload is hidden from the user.

lib/src/airship.dart
```diff
@@ -198,6 +198,24 @@ class Airship {
     return await _channel.invokeMethod('removeTags', tags);
   }
 
+  static Future<void> editChannelTagGroups(List<TagGroupOperation> operations) async {
```
rlepinski: This interface is still not a true editor:

Ideally its this:
```
Airship.editChannelTagGroups().addTag("group", "tag").remoteTag("group", "another tag").apply()
```

oristanovic: OK, so it looks like this now 
`Airship.editChannelTagGroups()..addTags("test_group", ["tag4","tag5"])..apply()`

rlepinski: Oh neat - https://dart.dev/guides/language/language-tour#cascade-notation-

PR #18, opened 2019-10-25
[MOBILE-594] Notification button support

oristanovic:
### What do these changes do?
Add custom notification button support

To allow for automatic loading of custom notification button categories

Created the UACustomNotificationCategories plist/xml file containing custom categories.
Tested manually on both iOS and Android.

### DIFF
--- a/ios/Classes/SwiftAirshipPlugin.swift
+++ b/ios/Classes/SwiftAirshipPlugin.swift
@@ -37,6 +37,8 @@ UADeepLinkDelegate, UAPushNotificationDelegate, UAInboxDelegate {
                                                selector: #selector(inboxUpdated),
                                                name: NSNotification.Name.UAInboxMessageListUpdated,
                                                object: nil)
+        
+        self.loadCustomNotificationCategories()
     }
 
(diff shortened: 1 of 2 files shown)


ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -280,6 +282,16 @@ UADeepLinkDelegate, UAPushNotificationDelegate, UAInboxDelegate {
             result(nil)
         }
     }
+    
+    private func loadCustomNotificationCategories() {
+        guard let categoriesPath = Bundle.main.path(forResource: "UACustomNotificationCategories", ofType: "plist") else { return }
+        let customNotificationCategories = UANotificationCategories.createCategories(fromFile: categoriesPath) as! Set<UANotificationCategory>
+    
+        if customNotificationCategories.count != 0 {
```
BrianBatchelder: I personally prefer "> 0" with counts. Just a suggestion. Feel free to leave as is.
```suggestion
        if customNotificationCategories.count > 0 {
```

rlepinski: No, it should be `!= 0`

PR #19, opened 2019-10-28
[MOBILE-945] fix Android launching OOTB Message Center in sample

crow:
### What do these changes do?
These changes allow onShowInboxMessage events to be correctly handled and display a custom message view instead of the OOTB message view.

To prevent the UI from transitioning to the OOTB message view and inbox when opening a message center message from a push.

Manually tested

https://urbanairship.slack.com/archives/G0B2N1MGX/p1572277892077200

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -6,6 +6,8 @@ import 'package:flutter/services.dart' show DeviceOrientation, SystemChrome;
 import 'package:airship_example/screens/home.dart';
 import 'package:airship_example/screens/settings.dart';
 import 'package:airship_example/screens/message_center.dart';
+import 'package:airship_example/screens/message_view.dart';
+
 import 'package:airship/airship.dart';
 
 // Supported deep links
@@ -30,6 +32,8 @@ class MyApp extends StatefulWidget {
 // SingleTickerProviderStateMixin is used for animation
 class _MyAppState extends State<MyApp> with SingleTickerProviderStateMixin {
   TabController controller;
+
+  final GlobalKey<NavigatorState> key = GlobalKey();
 
   @override
   void initState() {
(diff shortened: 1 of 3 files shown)


example/lib/main.dart
```diff
@@ -79,8 +83,18 @@ class _MyAppState extends State<MyApp> with SingleTickerProviderStateMixin {
...
-    Airship.onShowInboxMessage
-        .listen((messageId) => debugPrint('Show inbox message $messageId'));
+    Airship.onShowInboxMessage.listen((messageId){
+      Airship.inboxMessages.then((List<InboxMessage> messages) {
+        InboxMessage toShow = messages.firstWhere((thisMessage) =>
+        messageId == thisMessage.messageId,
+            orElse: () => null);
+
+        key.currentState.push(MaterialPageRoute<Null>(builder: (BuildContext context) {
+          return MessageView(message: toShow,);
```
marc-scig: what happens when toShow is null?

example/lib/screens/message_view.dart
```diff
@@ -4,9 +4,8 @@ import 'package:airship_example/styles.dart';
 
 class MessageView extends StatelessWidget {
   final InboxMessage message;
-  final updateParent;
 
-  MessageView({this.message, this.updateParent});
```
marc-scig: whats up with all these updateParent things?

marc-scig: (this is just for my better understanding, not saying I think the change is wrong)

crow: updateParent was basically a completion handler that would allow the message to make a function call in the parent context. The update state function it was calling was redundant - as the parent will update its state as it comes back into view.

example/lib/main.dart
crow: I don't think there's a very strong likelihood that the callback will be called with a message id that isn't in the list, but it would result in big red screen/NoSuchMethodError if it happened. Wrapped it in a check.

PR #20, opened 2019-10-30
[MOBILE-944] Update InboxMessageView

crow:
### What do these changes do?
Cleans up the InboxMessageView to be more like the react-native implementation. Implements this additional widget functionality in the sample.

These changes are necessary to improve the inbox message interface on the plugin for the first release.

Manually tested.

https://urbanairship.slack.com/files/U0298VBRH/FPN4X9TBM/update_inboxmessageview.mov

This gets the timeboxed work in place, but there is some follow up work to do here related to how the plugin interfaces platform-specific web view loading:
https://urbanairship.atlassian.net/browse/MOBILE-987

### DIFF
--- a/example/lib/screens/message_view.dart
+++ b/example/lib/screens/message_view.dart
@@ -33,7 +33,7 @@ class _MessageViewState extends State<MessageView>  {
 
           return Scaffold(
             appBar: AppBar(
-              title: Text("${message.title}"),
+              title: message != null ? Text("${message.title}") : Container(),
               backgroundColor: Styles.background,
             ),
             body: Stack(
@@ -41,9 +41,7 @@ class _MessageViewState extends State<MessageView>  {
                   isLoading
                       ? Center(child: CircularProgressIndicator())
                       : Container(),
-                  InboxMessageView(messageId: widget.messageId,
-                      onLoadStarted: onStarted,
-                      onLoadFinished: onLoadFinished)
+                  InboxMessageView(messageId: widget.messageId)
                 ]),
           );
         });
--- a/example/lib/screens/message_view.dart
+++ b/example/lib/screens/message_view.dart
@@ -3,9 +3,9 @@ import 'package:airship/airship.dart';
 import 'package:airship_example/styles.dart';
 
 class MessageView extends StatefulWidget {
-  final InboxMessage message;
+  final String messageId;
 
-  MessageView({this.message});
+  MessageView({this.messageId});
 
   @override
   _MessageViewState createState() => _MessageViewState();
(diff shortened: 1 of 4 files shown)


lib/src/inbox_message_view.dart
```diff
@@ -6,43 +6,51 @@ import 'package:flutter/widgets.dart';
 import 'airship.dart';
 
 class InboxMessageView extends StatefulWidget {
-  final InboxMessageViewCreatedCallback onViewCreated;
+  final InboxMessage message;
```
rlepinski: Require a message ID, not a message.

lib/src/inbox_message_view.dart
```diff
@@ -6,43 +6,51 @@ import 'package:flutter/widgets.dart';
...
 
-  InboxMessageViewController.init(int id) {
+  Future<void> onPlatformViewCreated(id) async {
     _channel =  new MethodChannel('com.airship.flutter/InboxMessageView_$id');
+    if (widget.onLoadStarted != null) {
+      widget.onLoadStarted();
+    }
+    loadMessage(widget.message).then((message) {
+      if (widget.onLoadFinished != null) {
+        widget.onLoadFinished();
```
rlepinski: This is not really the load finish we are looking for. This should happen after the message is fully loaded into the web view.

lib/src/inbox_message_view.dart
```diff
@@ -6,43 +6,51 @@ import 'package:flutter/widgets.dart';
...
+      }
+    });
   }
 
   Future<void> loadMessage(InboxMessage message) async {
     if (message == null) {
+      if (widget.onLoadError != null) {
+        widget.onLoadError(ArgumentError.notNull('message'));
```
rlepinski: Look at the react implementation. Should be from the native layer if it fails to load the message.

lib/src/inbox_message_view.dart
```diff
@@ -63,4 +71,12 @@ class _InboxMessageViewState extends State<InboxMessageView> {
 
     return new Text('$defaultTargetPlatform is not yet supported by this plugin');
   }
+
+  dispose() {
+    super.dispose();
+
+    if (widget.onClose != null) {
+      widget.onClose();
```
rlepinski: This is not that the widget closed, but if `UAirship.close()` is called from within the message view.

lib/src/inbox_message_view.dart
crow: added this

lib/src/inbox_message_view.dart
```diff
@@ -3,50 +3,21 @@ import 'dart:async';
...
-
-class _InboxMessageViewState extends State<InboxMessageView> {
-
   Future<void> onPlatformViewCreated(id) async {
-    if (widget.onViewCreated == null) {
-      return;
-    }
-    widget.onViewCreated(new InboxMessageViewController.init(id));
+    MethodChannel _channel =  new MethodChannel('com.airship.flutter/InboxMessageView_$id');
+    _channel.invokeMethod('loadMessage', messageId);
```
rlepinski: where is it handling the message ID changing?

crow: the inbox_message_view is a stateless and will build and reload the message when the messageId changes

example/lib/screens/message_view.dart
```diff
@@ -2,33 +2,60 @@ import 'package:flutter/material.dart';
...
+            list = List<InboxMessage>.from(snapshot.data);
+          }
+
+          message = list.firstWhere((thisMessage) =>
+          widget.messageId == thisMessage.messageId,
+              orElse: () => null);
+
+          return Scaffold(
+            appBar: AppBar(
+              title: message != null ? Text("${message.title}") : Container(),
```
BrianBatchelder: Is the view empty if the message is nil? Is there an alert like on iOS? Does it just show the inbox?

crow: The view body is empty until the web page request completes. The message property here is null until the message list future returns. In this case, we just show an invisible container until the title is ready to be displayed.

PR #21, opened 2019-11-01
[MOBILE-948] update iOS SDK to 12

crow:
### What do these changes do?
Updates the iOS SDK to 12 and performs requisite migration steps.

To have the latest iOS SDK.

Ran the sample, manually tested.

### DIFF
--- a/ios/airship.podspec
+++ b/ios/airship.podspec
@@ -3,7 +3,7 @@
 #
 Pod::Spec.new do |s|
   s.name             = 'airship'
-  s.version          = '0.0.1'
+  s.version          = '1.0.0'
   s.summary          = 'A new flutter plugin project.'
   s.description      = <<-DESC
 A new flutter plugin project.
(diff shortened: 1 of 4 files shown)


crow:
This is failing because the current airship-flutter iOS bitrise stack needs to be updated - which I'll do once this is in:https://github.com/urbanairship/airship-flutter/pull/20

ios/airship.podspec
```diff
@@ -16,7 +16,7 @@ A new flutter plugin project.
   s.public_header_files = 'Classes/**/*.h'
   s.dependency 'Flutter'
 
-  s.ios.deployment_target   = "10.0"
-  s.dependency              'UrbanAirship-iOS-SDK', '~> 11.0.0'
+  s.ios.deployment_target   = "11.0"
+  s.dependency              'UrbanAirship-iOS-SDK', '~> 12.0.0'
```
marc-scig: should we make this 12.0.1, now that it's out?

crow: yes we should

PR #27, opened 2020-01-04
[MOBILE-1049] update Android to 12.1.0, add channel attributes to and…

crow:
### What do these changes do?
Updates Android platform to SDK 12.1. and adds attribute support

Build/manually tested the Android example app.

image

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -9,7 +9,8 @@ import com.urbanairship.UAirship
 import com.urbanairship.analytics.CustomEvent
 import com.urbanairship.json.JsonMap
 import com.urbanairship.json.JsonValue
-import com.urbanairship.push.TagGroupsEditor
+import com.urbanairship.channel.TagGroupsEditor
+import com.urbanairship.channel.AttributeEditor
 import com.urbanairship.util.DateUtils
 import com.urbanairship.util.UAStringUtil
 import com.urbanairship.widget.UAWebView
@@ -30,6 +31,12 @@ private val TAG_OPERATION_TAGS = "tags"
 private val TAG_OPERATION_ADD = "add"
 private val TAG_OPERATION_REMOVE = "remove"
 private val TAG_OPERATION_SET = "set"
+
+private val ATTRIBUTE_MUTATION_TYPE = "action"
+private val ATTRIBUTE_MUTATION_KEY = "key"
+private val ATTRIBUTE_MUTATION_VALUE = "value"
+private val ATTRIBUTE_MUTATION_REMOVE = "remove"
+private val ATTRIBUTE_MUTATION_SET = "set"
 
 class InboxMessageViewFactory(private val registrar: Registrar) : PlatformViewFactory(StandardMessageCodec.INSTANCE) {
     override fun create(context: Context, viewId: Int, arguments: Any?): PlatformView {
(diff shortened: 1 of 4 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -31,6 +32,12 @@ private val TAG_OPERATION_ADD = "add"
 private val TAG_OPERATION_REMOVE = "remove"
 private val TAG_OPERATION_SET = "set"
 
+private val ATTRIBUTE_MUTATION_TYPE = "action"
```
BrianBatchelder: Would "ATTRIBUTE_MUTATION_ACTION" be clearer? If "ATTRIBUTE_MUTATION_TYPE" matches the other implementations, then don't worry about it.

lib/src/attribute_editor.dart
```diff
@@ -0,0 +1,34 @@
...
+  static const ATTRIBUTE_OPERATION_REMOVE = "remove";
+  static const ATTRIBUTE_OPERATION_SET = "set";
+  static const ATTRIBUTE_OPERATION_VALUE = "value";
+
+  final MethodChannel channel;
+  final String type;
+  final List<Map<String, dynamic>> operations;
+
+  AttributeEditor(String type, MethodChannel channel)
+      : this.type = type,
```
BrianBatchelder: Is this colon in the right place? Looks weird.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
marc-scig: it's probably my fault, I made a comment on either the react or cordova implementation about the name because I found the additional overloading of "action" confusing. but the string it maps to ("action") remained the same.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -242,6 +250,31 @@ class AirshipPlugin : MethodCallHandler {
...
     }
 
+    private fun editAttributes(call: MethodCall, result: Result) {
+        val mutations = call.arguments as ArrayList<Map<String, Any?>>
+        this.applyAttributesOperations(UAirship.shared().channel.editAttributes(), mutations)
+        result.success(null)
+    }
+
+    private fun applyAttributesOperations(editor: AttributeEditor, operations: ArrayList<Map<String, Any?>>) {
+        for (i in 0 until operations.size) {
```
marc-scig: this is not a critique, but since I am new to dart I did some googling and found that for looping over ArrayList you should also be able to say:

for (var operation in operations) {
}

and

operations.forEach((operation) {

)};

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
marc-scig: That's probably my fault, I made a previous comment on either the cordova or react implementation about how action was overloaded and confusing. but the string stayed the same. caveat emptor.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
crow: yeah not sure why I did this. Fixed with reasonable kotlin alternative

lib/src/attribute_editor.dart
crow: it's in the right place

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
crow: "action" is defined by the api, the name of this constant isn't that big of a deal, I think changing this to type was reasonable

PR #28, opened 2020-01-08
[MOBILE-1048] update iOS to 13.0.4

crow:
### What do these changes do?
Updates flutter iOS version to 13.0.4

Manually tested sample

flutter-ios-13 0 4

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -62,6 +62,9 @@ target 'Runner' do
   }
 end
 
+# Prevent Cocoapods from embedding a second Flutter framework and causing an error with the new Xcode build system.
+install! 'cocoapods', :disable_input_output_paths => true
+
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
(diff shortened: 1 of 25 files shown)


lib/airship.dart
```diff
@@ -1,4 +1,4 @@
-library airship;
+library airship_flutter;
```
rlepinski: is this the standard naming convention?

crow: yes

example/ios/Podfile
```diff
@@ -62,6 +62,9 @@ target 'Runner' do
   }
 end
 
+# Prevent Cocoapods from embedding a second Flutter framework and causing an error with the new Xcode build system.
```
BrianBatchelder: Is this a cocoa pods bug?

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -60,27 +61,36 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView {
...
+                             details:nil))
+       }
     }
 
     func view() -> UIView {
         return webView
     }
+
+    func close() {
+        // In the future we should send a close event to the dart layer
```
BrianBatchelder:
```suggestion
        // REVISIT: we should send a close event to the dart layer
```

example/ios/Podfile
crow: its actually a bug in xcode's new build system afaik

PR #30, opened 2020-01-10
[MOBILE-1040] add plugin tracking to ios and android

crow:
### What do these changes do?
Adds plugin tracking to ios and android. Follows yaml'ed version of the parsing pattern used for our react-native-module.

example:
https://github.com/urbanairship/react-native-module/pull/234/files

To send up the plugin version to analytics

Manually tested ios and android to ensure correct versions were being added to analytics.

image


android/build.gradle
```diff
@@ -1,6 +1,8 @@
 group 'com.airship.airship'
 version '1.0-SNAPSHOT'
 
+import org.yaml.snakeyaml.Yaml
```
rlepinski: 🐍

android/build.gradle
```diff
@@ -11,11 +13,21 @@ buildscript {
...
         classpath 'com.android.tools.build:gradle:3.5.3'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
+
+def getPluginVersion() {
+    Yaml yamlParser = new Yaml()
+
+    Map<String, Object> yaml = yamlParser.load(("../../pubspec.yaml" as File).text)
```
rlepinski: is the pubspec in the dart package?

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -4,18 +4,20 @@ import Airship
...
-    private let transactionIDKey = "transaction_id";
-    private let interactionIDKey = "interaction_id";
-    private let interactionTypeKey = "interaction_type";
-    private let tagOperationGroupName = "group";
-    private let tagOperationType = "operationType";
-    private let tagOperationTags = "tags";
-    private let tagOperationAdd = "add";
-    private let tagOperationRemove = "remove";
-    private let tagOperationSet = "set";
+    private let pluginVersion = "2.0.0"
```
rlepinski: we are setting this in two places

crow: nah, I moved this into its own class and forgot about this, removed

PR #34, opened 2020-01-31
[MOBILE-1225] Add screen tracking

oristanovic:
### What do these changes do?
Add screen tracking to flutter.

Ran the app and verified that the trackScreen method is called.


oristanovic:
> might want to include screen tracking calls in the flutter sample screens

added

PR #35, opened 2020-02-04
[MOBILE-1274] Update iOS to 13.1

oristanovic:
### What do these changes do?
Updates flutter iOS version to 13.1
Adds number attribute support

Manually tested the sample app

Capture d’écran 2020-02-05 à 11 52 44


oristanovic:
There are tickets for data collection support and the release itself. I don't see one for the number attributes so I added that here.

PR #37, opened 2020-02-06
[MOBILE-1272] add data collection enabled flag to flutter

crow:
### What do these changes do?
Add data collection and push token collection enablement to airship-flutter

To allow customers using the flutter integration to use these features.

Manually tested that the switching results in the correct method calls.

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -116,8 +116,10 @@ class AirshipPlugin : MethodCallHandler {
             "getInAppAutomationPaused" -> getInAppAutomationPaused(result)
             "enableChannelCreation" -> enableChannelCreation(result)
             "trackScreen" -> trackScreen(call, result)
-            "setPushTokenRegistrationEnabled" -> setPushTokenRegistrationEnabled(call, result)
+            "getDataCollectionEnabled" -> getDataCollectionEnabled(result)
+            "getPushTokenRegistrationEnabled" -> getPushTokenRegistrationEnabled(result)
             "setDataCollectionEnabled" -> setDataCollectionEnabled(call, result)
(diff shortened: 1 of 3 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -303,6 +306,16 @@ class AirshipPlugin : MethodCallHandler {
         result.success(true)
     }
 
+    private fun setDataCollectionEnabled(call: MethodCall, result: Result) {
+        UAirship.shared().isDataCollectionEnabled = call.arguments as Boolean
+        result.success(true)
+    }
+
+    private fun setPushTokenRegistrationEnabled(call: MethodCall, result: Result) {
+        UAirship.shared().pushManager.pushTokenRegistrationEnabled = (call.arguments as Boolean)
```
BrianBatchelder: Why is `call.arguments as Boolean` wrapped in parentheses here, but not a couple lines above?

lib/src/airship.dart
```diff
@@ -313,4 +313,20 @@ class Airship {
...
+    return await _channel.invokeMethod('setDataCollectionEnabled', enabled);
+  }
+
+  static Future<bool> setPushTokenRegistrationEnabled(bool enabled) async {
+    if (enabled == null) {
+      throw ArgumentError.notNull('enabled');
+    }
+
+    return await _channel.invokeMethod('setPushTokenRegistrationEnabled', enabled);
+  }
```
BrianBatchelder: FWIW, these two methods are implemented in the opposite order in the platform-dependent code.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
crow: I converted these from their setters, because that's what kotlin likes, no strong reason

lib/src/airship.dart
crow: reordered these to be consisten

PR #38, opened 2020-02-07
[MOBILE-1286] Move Flutter deploy to Github Actions

crow:
### What do these changes do?
Adds flutter deploy job to github actions.

To allow flutter deploys from github actions.

The build/ios/android jobs are all verified and currently in use. Not able to verify publish and finish jobs yet, will have to run them. Verified pub-dev credentials and slack hook.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,65 @@
+  
+name: Release
+
+on:
+  push:
+    tags:
+      - "[0-9]+.[0-9]+.[0-9]+*"
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: "12.x"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+      - uses: subosito/flutter-action@v1
+        with:
+          channel: "stable"
+      - name: Run CI
+        run: bash ./scripts/run_ci_tasks.sh -a
+  ios:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
```
marc-scig: is this copied from the android section? what do we need to set up java for here?

crow: This is just a direct port of the CI jobs - but yeah, this looks weird. I don't think analyze or ios need java set up, removed.

marc-scig: Ok, let's leave it for now then. We can revisit it later.

PR #40, opened 2020-02-07
[MOBILE-987] Handle webview delegate callbacks

oristanovic:
### What do these changes do?
Handle WKWebView navigation delegate callbacks. I tried to follow React Native implementation as much as possible, let me know if I missed something

Manual tests in the sample app


example/lib/screens/message_view.dart
```diff
@@ -59,9 +82,34 @@ class _MessageViewState extends State<MessageView>  {
...
     });
   }
+
+  onLoadError(PlatformException e) {
+    setState(() {
+      isLoading = false;
+      showDialog(
+          context: context,
+          builder: (context) => AlertDialog(
+            title: Text(e.message != null ? e.message : ""),
```
BrianBatchelder: Should you default the message to something? Otherwise, the user could see a dialog with no title and no content.

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -19,9 +19,13 @@ class AirshipInboxMessageViewFactory : NSObject, FlutterPlatformViewFactory {
 }
 
 class AirshipInboxMessageView : NSObject, FlutterPlatformView, UANativeBridgeDelegate, WKNavigationDelegate {
+    let loadStartedEvent = AirshipWebviewLoadStartedEvent()
+    let loadFinishedEvent = AirshipWebviewLoadFinishedEvent()
+    let closeEvent = AirshipWebviewClosedEvent()
```
BrianBatchelder: Maybe `closedEvent` so the naming is consistent?

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -91,6 +98,40 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, UANativeBridgeDel
...
+    }
+    
+    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        AirshipEventManager.shared.notify(loadFinishedEvent)
+    }
+    
+    func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
+        webviewResult(FlutterError(code:"MessageLoadFailed",
+        message:"Unable to load message",
+        details:error.localizedDescription))
```
BrianBatchelder: Indentation

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -60,6 +65,8 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, UANativeBridgeDel
     }
 
     private func loadMessage(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        webviewResult = result
+        AirshipEventManager.shared.notify(loadStartedEvent)
```
BrianBatchelder: Do you need to emit a `loadFinishedEvent` in the "InvalidMessage" else block below (lines 89-93)?

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -91,6 +98,40 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, UANativeBridgeDel
...
+                decisionHandler(.cancel)
+                if (response.statusCode == 410) {
+                    webviewResult(FlutterError(code:"MessageLoadFailed",
+                           message:"Unable to load message",
+                           details:"Message not available"))
+                } else {
+                    webviewResult(FlutterError(code:"MessageLoadFailed",
+                           message:"Unable to load message",
+                           details:"Message load failed"))
+                }
```
BrianBatchelder: Do you need to emit a `loadFinishedEvent` here?

ios/Classes/AirshipInboxMessageView.swift
BrianBatchelder: Also, do you need to emit a `loadFinishedEvent` here?

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -91,6 +98,40 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, UANativeBridgeDel
...
+    }
+    
+    func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
+        webviewResult(FlutterError(code:"MessageLoadFailed",
+        message:"Unable to load message",
+        details:error.localizedDescription))
+    }
+    
+    func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
+        webView.navigationDelegate?.webView?(webView, didFail: navigation, withError: error)
```
BrianBatchelder: And do you need to emit a `loadFinishedEvent` here? (sorry for all of the duplicate comments)

ios/Classes/Events/WebviewClosedEvent.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
```
BrianBatchelder: This file should be named `AirshipWebviewClosedEvent.swift`.

ios/Classes/Events/WebviewLoadFinishedEvent.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
```
BrianBatchelder: This file should be named `AirshipWebviewLoadFinishedEvent .swift`.

ios/Classes/AirshipInboxMessageView.swift
oristanovic: The way I see it is that loadFinished means load _successfully_ finished so it wouldn't make sense to call it in error scenarios.

BrianBatchelder: I'm concerned when I see notifications that don't always "close the loop". I can imagine an app depending on seeing `loadStarted` and `loadFinished` to indicate the message has loaded. But I also understand your desire to have `loadFinished` indicate success. Perhaps we need a `loadFailed`? Otherwise, the app has no way of knowing that except for setting an arbitrary timeout, right?

oristanovic: `onLoadError` is called in the app through the FlutterResult callback:
```
webviewResult(FlutterError(code:"MessageLoadFailed",
                           message:"Unable to load message",
                           details:"Message load failed"))
```

oristanovic:
> Some minor stuff. Also, is there Android work to be done for this?

Actually yes, added the missing Android work. For some reason I thought this was an iOS only ticket.

ios/Classes/AirshipInboxMessageView.swift
BrianBatchelder: Oh, nice. I missed that.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -49,9 +56,35 @@ class InboxMessageViewFactory(private val registrar: Registrar) : PlatformViewFa
...
 
+    lateinit private var webviewResult:Result
+
     private val webView: UAWebView by lazy {
         val view = UAWebView(context)
-        view.webViewClient = UAWebViewClient()
+        view.webViewClient = object: UAWebViewClient() {
+            override fun onPageStarted(view: WebView, url: String?, favicon: Bitmap?) {
+                super.onPageStarted(view, url, favicon)
+                EventManager.shared.notifyEvent(WebviewLoadStartedEvent())
```
rlepinski: I dont think you want to use `EventManager`,  instead these events should be sent through the `Registrar`/`Channel` in the `InboxMessageViewFactory`

oristanovic: Ok, changed the approach

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -63,17 +58,17 @@ class FlutterInboxMessageView(private var context: Context) : PlatformView, Meth
...
             override fun onPageFinished(view: WebView?, url: String?) {
                 super.onPageFinished(view, url)
-                EventManager.shared.notifyEvent(WebviewLoadFinishedEvent())
+                channel.invokeMethod("onLoadFinished", null)
             }
 
             override fun onClose(webView: WebView) {
                 super.onClose(webView)
-                EventManager.shared.notifyEvent(WebviewClosedEvent())
+                channel.invokeMethod("onClose", null)
```
rlepinski: Where is onLoadError being sent?

example/lib/screens/message_view.dart
```diff
@@ -82,28 +60,41 @@ class _MessageViewState extends State<MessageView>  {
     });
   }
 
-  onLoadStarted() {
+  onLoadEventReceived(String method, PlatformException e) {
     setState(() {
-      isLoading = true;
+      switch (method) {
+        case 'onLoadStarted':
```
rlepinski: I think we need to make separate handlers for each type of event in the inbox_message_view.dart that hides the raw event data so we can do extra parsing on the data if needed. Sorta like we do for the other events.

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
oristanovic: in onReceivedError through FlutterResult

example/lib/screens/message_view.dart
oristanovic: ok

oristanovic: actually just to be sure we're on the same page, you're talking abut FlutterMethodChannel or FlutterEventChannel? Because here I switched to the method channel

rlepinski: What I think we want to break up the handlers:

```
InboxMessageView(messageId: widget.messageId, 
onLoad:handleLoad, 
onError:handleError,
onClose:handleClose
etc...)
```

Make all the handlers optional.

rlepinski:
Looks great but I think the methods names should be reversed. InboxMessageView should allow setting `onLoad` and the example method should be `handleLoad`

PR #41, opened 2020-02-19
Fix release

crow:
### What do these changes do?
These changes refactor the airship class to airship_flutter to resolve a warning on publish. This allows the plugin to pass the publish dry run that our github actions tool runs before force publishing.

To make it possible to release using the current github actions job

Ran both samples, ran and passed publish dry run

image
image

image


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Version 3.0.0 - February 19, 2020
+================================
+- Refactored airship class name to airship_flutter to resolve publish warning
```
rlepinski: Mention what customers will have to do for the update

marc-scig: Do you mean "file name" here? Seems like the class is still Airship.

marc-scig: Did we still need to mention what customers need to do in order to update? I assume we at least want to say something about changing import statements.

crow: it's the dart interface, I guess it's not technically a class

PR #42, opened 2020-02-28
Mobile 1320: Generate API docs for flutter

khmMouna:
### What do these changes do?
Generate API docs for flutter

I updated the release workflow. I added a new step `upload-docs:`


BrianBatchelder:
Mouna, could you add some text to the title? You can get it from the title of the ticket. In this case, maybe something like `Mobile 1320: Generate API docs for flutter`.

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+#####################################################
+# This script is used for Continuous Integration
+#
+# Run locally to verify before committing your code.
+#
+# Options:
+#   -g to generate documentations.
```
BrianBatchelder:
```suggestion
#   -g to generate documentation.
```

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
...
+set -o pipefail
+set -e
+
+# get platforms
+GENERATE=false
+UPLOAD=false
+
+while true; do
+  case "$1" in
+    -g  ) GENERATE=true;UPLOAD=false;;
```
BrianBatchelder:
```suggestion
    -g  ) GENERATE=true;;
```
No reason generate and upload couldn't both be run in the same script execution, is there?

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
...
+    -u  ) UPLOAD=true;GENERATE=false;;
```
BrianBatchelder:
```suggestion
    -u  ) UPLOAD=true;;
```

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    if [ ! -d "$2" ]; then
+        echo "Missing docs $2"
+        exit 1
+    fi
+
+    ROOT_PATH=`dirname "${0}"`/..
+    TAR_NAME="$1.tar.gz"
+
+    cd doc/
```
BrianBatchelder:
```suggestion
    cd $2
```

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
...
+    tar -czf $TAR_NAME *
+    cd -
+
+    gsutil cp $ROOT_PATH/doc/$TAR_NAME gs://ua-web-ci-prod-docs-transfer/libraries/flutter/$TAR_NAME
```
BrianBatchelder:
```suggestion
    gsutil cp $ROOT_PATH/$2/$TAR_NAME gs://ua-web-ci-prod-docs-transfer/libraries/flutter/$TAR_NAME
```

scripts/docs.sh
```diff
@@ -0,0 +1,58 @@
...
+    -g  ) GENERATE=true;UPLOAD=false;;
+    -u  ) UPLOAD=true;GENERATE=false;;
+    *   ) break ;;
+  esac
+  shift
+done
+
+# Generate doc
+if $GENERATE; then
+    $FLUTTER_ROOT/bin/cache/dart-sdk/bin/dartdoc  --exclude 'dart:async,dart:collection,dart:convert,dart:core,dart:developer,dart:ffi,dart:html,dart:io,dart:isolate,dart:js,dart:js_util,dart:math,dart:typed_data,dart:ui'
```
BrianBatchelder:
```suggestion
  $FLUTTER_ROOT/bin/cache/dart-sdk/bin/dartdoc --exclude 'dart:async,dart:collection,dart:convert,dart:core,dart:developer,dart:ffi,dart:html,dart:io,dart:isolate,dart:js,dart:js_util,dart:math,dart:typed_data,dart:ui'
```

scripts/docs.sh
khmMouna: Yes, you are right, we can generate and upload doc with the same script

.github/workflows/release.yml
```diff
@@ -42,9 +42,47 @@ jobs:
           channel: "stable"
       - name: Run CI
         run: bash ./scripts/run_ci_tasks.sh -i
+  upload-docs:
+    runs-on: macos-latest
```
rlepinski: do we need this on macos-latest?

.github/workflows/release.yml
```diff
@@ -42,9 +42,47 @@ jobs:
...
+    needs: [analyze, android, ios]
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '12.x'
+      - name: Flutter action
+        uses: subosito/flutter-action@v1
+        with:
+          flutter-version: '1.12.13+hotfix.8'
```
rlepinski: why is this version different?

.github/workflows/release.yml
khmMouna: I chose mac to have the same environment for the test but we can use ubuntu-latest also.

.github/workflows/release.yml
khmMouna: It's the last stable version. I fixed the version because I didn't find another way to get the path of the FLUTTER_ROOT. 
 `FLUTTER_ROOT: /Users/runner/hostedtoolcache/flutter/1.12.13-hotfix.8-stable/x64`

PR #43, opened 2020-03-05
[MOBILE-1345] Add github release

rlepinski:
Adds creating a github release during our release flow. Reformats the changelog to make it easier to pull the release notes. Simplified the release workflow and added a check version step.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,34 +6,12 @@ on:
       - "[0-9]+.[0-9]+.[0-9]+*"
 
 jobs:
-  analyze:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -6,94 +6,76 @@ on:
...
         with:
           java-version: "12.x"
       - uses: subosito/flutter-action@v1
         with:
           channel: "stable"
       - name: Run CI
-        run: bash ./scripts/run_ci_tasks.sh -i
-  upload-docs:
+        run: bash ./scripts/run_ci_tasks.sh
+  release:
```
BrianBatchelder: It looks like this step is called `deploy` in the react CI, but `release` here and in Android. Would it be better to be consistent?

PR #47, opened 2020-04-30
Adds updateRegistration() support to the Flutter API

imjacobclark:
Co-authored-by: Jacob Clark <jacob.jh.clark@googlemail.com>
Co-authored-by: Chris Grounds <chrisbacon2009@hotmail.com>

### What do these changes do?
Exposes the updateRegistration API to the Flutter client to enable apps to immediately register their users preferences when toggling notifications on or off.

To enable Flutter developers to choose when their notification permissions are synced to prevent users receiving notifications they have opted out of.

Manually integrated the package into a test Flutter app in Android and iOS.

N/A

N/A

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -129,6 +129,7 @@ class AirshipPlugin : MethodCallHandler {
             "clearNotifications" -> clearNotifications(result)
             "getActiveNotifications" -> getActiveNotifications(result)
             "addTags" -> addTags(call, result)
+            "updateRegistration" -> updateRegistration(call, result)
             "addEvent" -> addEvent(call, result)
             "removeTags" -> removeTags(call, result)
             "getTags" -> getTags(result)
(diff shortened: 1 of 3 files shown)


rlepinski:
@imjacobclark Thanks for the PR.

What methods are you calling in the plugin which require `updateRegistration` ? Ideally that is handled for you by the plugin and you don't have to worry about when/how the Airship SDK works.

imjacobclark:
> @imjacobclark Thanks for the PR.
> 
> 
> 
> What methods are you calling in the plugin which require `updateRegistration` ? Ideally that is handled for you by the plugin and you don't have to worry about when/how the Airship SDK works. 

Hey @rlepinski - we’re hitting the notificationsEnabled API on iOS and Android. 

It works fine on Android however there is a significant delay on iOS to the preferences being updated.

rlepinski:
@imjacobclark Thanks for the info, before we expose this method I would like to see about fixing the enable notification method on iOS.

imjacobclark:
> @imjacobclark Thanks for the info, before we expose this method I would like to see about fixing the enable notification method on iOS. 

Okay thanks. Do you have an ETA on when you may have a fix for that? We could use a forked version of the package for now in our codebase but that’s not ideal really, thanks.

rlepinski:
I don't have a timeline yet but i will have someone take a deeper look at it today. We are usually pretty quick about fixing issues.

rlepinski:
@imjacobclark We found and fixed the issue in the underlying iOS SDK. We will release an update in the next week or so with the fix.

imjacobclark:
> @imjacobclark We found and fixed the issue in the underlying iOS SDK. We will release an update in the next week or so with the fix.

Ace! Thanks for the quick resolution!

rlepinski:
Should be resolved in master if you want to use that as your branch, working on a release today.

rlepinski:
Released in 3.0.2

PR #53, opened 2020-05-20
[MOBILE-1565] Add named user attributes

oristanovic:
### What do these changes do?
Add named user attribute support to Flutter.

To allow customers using flutter to use the feature.

Manual tests.

### DIFF
--- a/ios/Classes/SwiftAirshipPlugin.swift
+++ b/ios/Classes/SwiftAirshipPlugin.swift
@@ -121,6 +121,8 @@ UADeepLinkDelegate, UAPushNotificationDelegate {
             getTags(call, result: result)
         case "editAttributes":
             editAttributes(call, result: result)
+        case "editNamedUserAttributes":
+            editNamedUserAttributes(call, result: result)
         case "editNamedUserTagGroups":
             editNamedUserTagGroups(call, result: result)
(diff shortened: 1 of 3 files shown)


ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -289,28 +291,22 @@ UADeepLinkDelegate, UAPushNotificationDelegate {
 
     private func editAttributes(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
```
BrianBatchelder, line 292: You might rename private methods to `editChannelAttributes` throughout, to be symmetric with `editNamedUserAttributes`. Probably not worth a change to the public I/F, but maybe even there.

lib/src/airship_flutter.dart
```diff
@@ -215,6 +215,10 @@ class Airship {
     return AttributeEditor('editAttributes', _channel);
   }
 
+  static AttributeEditor editNamedUserAttributes() {
```
rlepinski, line 218: could we deprecate `editAttributes` and add new `editChannelAttributes`

oristanovic: done

PR #54, opened 2020-05-20
[MOBILE-1563] - Remove property restrictions for  custom events in flutter

khmMouna:
### What do these changes do?
Update the custom event property to `setProperties(Map<String, dynamic>)` and `addProperty(String, dyamic)`. 

Using the example app.
example of code: 
```
    CustomEvent event = CustomEvent("Event name", 123);
    event.addProperty("test", true);
    event.addProperty("First name", true);
    event.addProperty("Last name", "lastname");
```

```
    var properties = new Map<String, dynamic>();
    properties["First name"] = "Me";
    properties["age"] = 30;
    event.setProperties(properties);
    Airship.addEvent(event);
```

### DIFF
--- a/lib/src/custom_event.dart
+++ b/lib/src/custom_event.dart
@@ -35,7 +35,7 @@ class CustomEvent {
   }
 
   ///
-  /// Sets a custom event properties.
+  /// Sets custom event properties.
   ///
   /// @param properties The custom event properties.
   ///
--- a/lib/src/custom_event.dart
+++ b/lib/src/custom_event.dart
@@ -19,45 +19,31 @@ class CustomEvent {
         this.value = value;
 
   ///
-  /// Sets a custom BOOL property.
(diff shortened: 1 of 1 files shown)


lib/src/custom_event.dart
```diff
@@ -19,45 +19,31 @@ class CustomEvent {
...
+    if (value == null) {
+      throw ArgumentError.notNull('value');
+    }
     _properties[key] = value;
   }
 
   ///
-  /// Sets a custom String property. The value's length must not exceed 255 characters
-  /// or it will invalidate the event.
+  /// Sets a custom event properties.
```
BrianBatchelder, line 38:
```suggestion
  /// Sets custom event properties.
```

khmMouna:
@BrianBatchelder I confirm that it works as expected with complex type.

Issue #56, opened 2020-06-28
Add Flutter Module to a Native App


rlecheta:
Hi, I'm doing this for a app, so I can migrate few modules of my app using flutter.
https://flutter.dev/docs/development/add-to-app

But the airship-flutter plugin doesn't work as a module because it was done with the old flutter plugin system and it should be migrated:
https://flutter.dev/docs/development/packages-and-plugins/plugin-api-migration

This is the error, after create a Flutter Module with the airship-flutter plugin:

`The plugin `airship_flutter` is built using an older version of the Android plugin API which assumes that it's running in a full-Flutter environment. It may have undefined behaviors when Flutter is integrated into an existing app as a module.
The plugin can be updated to the v2 Android Plugin APIs by following https://flutter.dev/go/android-plugin-migration.`

I think this is a easy fix, looking forward to see it, thank you.

rlepinski:
Thanks for reporting this, we will update it in our next release.

PR #57, opened 2020-07-27
[MOBILE-1739] Update Android flutter plugin to new api version

Apekka:
### What do these changes do?
These changes upgrade the Flutter Android plugin to new API version

To solve an issue reported by a customer
https://github.com/urbanairship/airship-flutter/issues/56

By running the app, and testing the basic functionalities of Airship.


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -110,7 +116,12 @@ class FlutterInboxMessageView(private var context: Context, channel: MethodChann
     }
 }
 
-class AirshipPlugin : MethodCallHandler {
+class AirshipPlugin : MethodCallHandler, FlutterPlugin, ActivityAware, ServiceAware {
```
rlepinski, line 119: do we need to implement ActivityAware and ServiceAware interfaces?

Apekka: In the Flutter docs, it says :
```
(Optional) If your plugin needs an Activity reference, also implement the ActivityAware interface.

(Optional) If your plugin is expected to be held in a background Service at any point in time, implement the ServiceAware interface.
```

And if I'm not wrong, I guess the Plugin interacts with the Activity, and is expected to be held in a background Service.
However, looks like everything is working fine even when I remove ActivityAware and ServiceAware. So I guess I will remove them

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -110,18 +112,32 @@ class FlutterInboxMessageView(private var context: Context, channel: MethodChann
...
 
+    override fun onAttachedToEngine(binding: FlutterPlugin.FlutterPluginBinding) {
+        channel = MethodChannel(binding.flutterEngine.dartExecutor, "com.airship.flutter/airship")
+        binding.platformViewRegistry.registerViewFactory("com.airship.flutter/InboxMessageView", InboxMessageViewFactory(binding))
+        channel.setMethodCallHandler(this)
+        context = binding.applicationContext
+    }
+
+    override fun onDetachedFromEngine(binding: FlutterPlugin.FlutterPluginBinding) {
+        channel.setMethodCallHandler(null)
```
marc-scig, line 138: should we also release the other references here (channel, context)?

Apekka: Yeah I released the channel here, but I don't think that releasing the context is needed.
However, I may be wrong

PR #70, opened 2020-12-04
[CUSTENG-3584] Add support for refreshing message center messages in flutter

devinsmythe:
### What do these changes do?

These changes should allow customers using the flutter SDK to manually refresh the inbox. Currently refreshes only happen when opening the app or when a push is received with a message center message.

A customer wants to be able to refresh the inbox inside the app.
https://urbanairship.atlassian.net/browse/CUSTENG-3584

I tested the changes on an android device.
**I have not tested the changes on iOS**, I could use some help testing my iOS changes out!

iOS still needs to be tested. I'm not 100% sure I called the correct iOS method.

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -163,7 +163,7 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
             "setNamedUser" -> setNamedUser(call, result)
             "getNamedUser" -> getNamedUser(result)
             "getInboxMessages" -> getInboxMessages(result)
-            "fetchMessages" -> fetchMessages(result)
+            "refreshInbox" -> refreshInbox(result)
             "markInboxMessageRead" -> markInboxMessageRead(call, result)
             "deleteInboxMessage" -> deleteInboxMessage(call, result)
             "setInAppAutomationPaused" -> setInAppAutomationPaused(call, result)
@@ -203,9 +203,10 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
         result.success(messages)
     }
 
-    private fun fetchMessages(result: Result) {
-        MessageCenter.shared().inbox.fetchMessages()
-        result.success(null)
+    private fun refreshInbox(result: Result) {
+        MessageCenter.shared().inbox.fetchMessages { sucess ->
+            result.sucess(sucess)
+        }
     }
 
     private fun markInboxMessageRead(call: MethodCall, result: Result) {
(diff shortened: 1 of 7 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -202,6 +203,11 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
         result.success(messages)
     }
 
+    private fun fetchMessages(result: Result) {
+        MessageCenter.shared().inbox.fetchMessages()
```
rlepinski, line 207: wait to set the result until you get a callback:
```
MessageCenter.shared().inbox.fetchMessages { success -> 
  result.success(success)            
}

File: ios/Classes/SwiftAirshipPlugin.swift [hunk 2]
```diff
@@ -392,6 +392,10 @@ UADeepLinkDelegate, UAPushNotificationDelegate {
         }
     }
 
+    private func fetchMessages(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        UAMessageCenter.shared().messageList.refreshInboxWithCompletionHandler()
```
rlepinski, line 396: need to set the result as well, ideally in the completion handler

lib/src/airship_flutter.dart
```diff
@@ -250,6 +250,10 @@ class Airship {
     return await _channel.invokeMethod('deleteInboxMessage', message.messageId);
   }
 
+  static Future<void> refreshInbox() async {
```
rlepinski, line 253: can be updated to Future<bool> if you do the chaining right with the completion handlers

lib/src/airship_flutter.dart
```diff
@@ -250,6 +250,10 @@ class Airship {
...
+    return await _channel.invokeMethod("fetchMessages");
```
rlepinski, line 254: you should keep method names the same so its easier to trace, `refreshInbox` here and in the native layer

rlepinski:
Ill look into the failing CI.

The suggestions above will make it easier to implement a refresh loading indicator in the app since they will be notified with the refresh is finished using the future.

devinsmythe:
@rlepinski having issues with the swift implementation. Can you take a look at what I'm doing wrong?

devinsmythe:
Hopefully I got the future stuff working correctly. I tested it on android with the sample and I was able to retrieve the future value doing this so seems like it's working:
`Airship.refreshInbox().then((value) => print("Future value: " + value.toString()));`

oristanovic (approved):

Apekka (approved):

PR #80, opened 2021-04-12
[MOBILE-2195] Add null safety annotations

Ulrico972:
### What do these changes do?
Migrate the plugin to use non-nullable types and Add null safety annotations.

Make the plugin support null safety.

Build, run, send some message in the message center and check tags.
But I think some QA tests would be valuable just in case.

I used the migration tool then checked and manually adapted the suggestions.

### DIFF
--- a/lib/src/inbox_message_view.dart
+++ b/lib/src/inbox_message_view.dart
@@ -22,20 +22,28 @@ class InboxMessageView extends StatelessWidget {
     MethodChannel _channel = new MethodChannel('com.airship.flutter/InboxMessageView_$id');
     _channel.setMethodCallHandler(methodCallHandler);
     _channel.invokeMethod('loadMessage', messageId).catchError( (error) {
-      onLoadError!(error);
+      if (onLoadError != null) {
+        onLoadError!(error);
+      }
     });
   }
 
   Future<void> methodCallHandler(MethodCall call) async {
     switch (call.method) {
       case 'onLoadStarted':
-        onLoadStarted!();
+        if (onLoadStarted != null) {
+          onLoadStarted!();
+        }
         break;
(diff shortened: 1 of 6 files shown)


lib/src/inbox_message_view.dart
```diff
@@ -8,34 +8,34 @@ import 'package:flutter/widgets.dart';
...
-    @required this.messageId, this.onLoadStarted, this.onLoadFinished, this.onLoadError, this.onClose
+    required this.messageId, this.onLoadStarted, this.onLoadFinished, this.onLoadError, this.onClose
   });
 
   Future<void> onPlatformViewCreated(id) async {
     MethodChannel _channel = new MethodChannel('com.airship.flutter/InboxMessageView_$id');
     _channel.setMethodCallHandler(methodCallHandler);
     _channel.invokeMethod('loadMessage', messageId).catchError( (error) {
-      onLoadError(error);
+      onLoadError!(error);
```
marc-scig, line 25: Are we certain these Functions will exist when we call them? My impression reading this is that everything except the message ID is optional and the

lib/src/inbox_message_view.dart
```diff
@@ -8,34 +8,34 @@ import 'package:flutter/widgets.dart';
...
     });
   }
 
   Future<void> methodCallHandler(MethodCall call) async {
     switch (call.method) {
       case 'onLoadStarted':
-        onLoadStarted();
+        onLoadStarted!();
```
marc-scig, line 32: same comment as above, I think we may not want to use the null assertion operator here unless I'm missing something.

marc-scig (approved):

oristanovic (approved):

PR #89, opened 2021-08-03
[MOBILE-2426] Expose badge methods

oristanovic:
Exposes resetBadge, setBadge, setAutoBadgeEnabled, isAutobadgeEnabled.
Tested on iOS sample app.

### DIFF
--- a/lib/src/airship_flutter.dart
+++ b/lib/src/airship_flutter.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 import 'package:airship_flutter/src/attribute_editor.dart';
 import 'package:flutter/services.dart';
 import 'custom_event.dart';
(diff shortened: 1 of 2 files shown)


lib/src/airship_flutter.dart
```diff
@@ -310,4 +311,36 @@ class Airship {
...
+
+  static Future<bool?> get isAutoBadgeEnabled async {
+    if (Platform.isIOS) {
+      return await _channel.invokeMethod('isAutoBadgeEnabled');
+    } else if (Platform.isAndroid) {
+      return false;
+    }
+  }
+
+  static Future<bool?> setAutoBadgeEnabled(bool enabled) async {
```
rlepinski, line 323: why is this a bool? return type

lib/src/airship_flutter.dart
```diff
@@ -310,4 +311,36 @@ class Airship {
   static Future<bool?> setPushTokenRegistrationEnabled(bool enabled) async {
     return await _channel.invokeMethod('setPushTokenRegistrationEnabled', enabled);
   }
+
+  static Future<bool?> get isAutoBadgeEnabled async {
```
rlepinski, line 315: should it just be `bool`?

lib/src/airship_flutter.dart
```diff
@@ -310,4 +311,36 @@ class Airship {
...
+    } else if (Platform.isAndroid) {
+      return false;
+    }
+  }
+
+  static Future<void> setBadge(int badge) async {
+    if (Platform.isIOS) {
+      return await _channel.invokeMethod('setBadge', badge);
+    } else if (Platform.isAndroid) {
+      return;
```
rlepinski, line 335: do you need to return an empty future?

lib/src/airship_flutter.dart
```diff
@@ -310,4 +311,30 @@ class Airship {
...
+  static Future<bool> isAutoBadgeEnabled() async {
+    var isAutoBadgeEnabled = false;
+    if (Platform.isIOS) {
+      isAutoBadgeEnabled = await _channel.invokeMethod('isAutoBadgeEnabled');
+    }
+    return isAutoBadgeEnabled;
+  }
+
+  static Future<void> setAutoBadgeEnabled(bool enabled) async {
+    if (Platform.isIOS) {
```
marc-scig, line 324: it looks like these set methods will never complete on android if someone is waiting on the future. we may want to resolve the future immediately to prevent that when not on iOS.

marc-scig (approved):

Apekka (approved):

PR #91, opened 2021-09-20
[MOBILE-2525] Release 4.4.0

Ulrico972:
### What do these changes do?
Update SDKs version

For the release 4.4.0

Build and run.

I had to update the flutter version used for the CI due to issues with dartdoc. So this use the version 3.1.0 of dartdoc and 2.14.0 of dart sdk.
I updated the gitignore to take the content of .idea instead of the directory itself to solve an issue with the android/.idea/vcs.xml file that was ignored but needed by pub to be published. I wanted to use a pubignore file but no matter what I tried, it just failed.
Last thing, I added the NDK dir path in the local.properties because it's actually using the lasts versions by default which doesn't contains all the ABI in the toolchains. That fixes the "No toolchains found in the NDK toolchains folder for ABI with prefix: arm-linux-androideabi" issue.


.github/workflows/ci.yml
```diff
@@ -27,9 +28,10 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: "12.x"
-      - uses: subosito/flutter-action@123325ea84b0d4c11e96eb9899090b697bf51bed
+      - uses: subosito/flutter-action@v1
         with:
-          flutter-version: "2.0.3"
+          channel: 'stable'
```
oristanovic, line 33: does this need to be explicit ?

oristanovic (approved):

.github/workflows/ci.yml
Ulrico972: We could skip the stable mention actually. But the flutter version should be fix to avoid the dartdoc and dart-sdk versions to change

Ulrico972:
Ok, it seems like the NDK issue is not actually resolved yet... 😑

rlepinski (approved):
Assuming you can get CI working

PR #92, opened 2021-10-01
[MOBILE-2527] Update to latest SDKs

oristanovic:
Updates to iOS SDK 16 and Android SDK 15


oristanovic:
> @oristanovic This needs to be rebased on main, then it should resolve your CI issues

it's ready now

android/build.gradle
```diff
@@ -4,7 +4,7 @@ version '1.0-SNAPSHOT'
 buildscript {
     ext.kotlin_version = '1.4.21'
     ext.coroutine_version = '1.3.0'
-    ext.airship_version = '14.4.3'
+    ext.airship_version = '15.0.0'
```
rlepinski, line 7: Need 15.1.0

rlepinski (approved):
We will have a patch release tomorrow, so we can bump versions before doing a release. This is fine though to merge.

ios/Classes/Events/AirshipPushReceivedEvent.swift
```diff
@@ -16,20 +16,7 @@ class AirshipPushReceivedEvent : AirshipEvent {
...
-                notificaitonPayload["title"] = notificationContent.alertTitle
-                notificaitonPayload["notification_id"] = notification.request.identifier
-
-                var extras = notificationContent.notificationInfo
-                extras["_"] = nil
-                extras["aps"] = nil
-                notificaitonPayload["extras"] = extras
-                payload["notification"] = notificaitonPayload
-            }
-
```
khmMouna, line 32: I have the same issue with react native. Some informations are lost with the new update. @rlepinski thinks that we need to keep the current push body to avoid breaking integrations. I don't think it's possible, mybe we just need to add document on that. 
I think also you need to exhibit the extras only not all the payload. Not sure about that.

khmMouna (commented):
Aren't we going to add 
- `getConfiguration`
- `editSubscriptionLists `
- `setUseCustomPreferenceCenterUi`
- `addPrefereceCenterOpenListener`
To be synchronised with other plugins ?

rlepinski:
I am ok holding off on those, but it should be a fast follow up.

PR #100, opened 2021-11-16
Mobile 2675

oristanovic:


ios/Classes/AirshipPlugin.m
```diff
@@ -3,6 +3,16 @@
...
+    [SwiftAirshipPlugin registerWithRegistrar:registrar];
+}
+
++ (void)load {
+    NSNotificationCenter *center = [NSNotificationCenter defaultCenter];
+       [center addObserverForName:UIApplicationDidFinishLaunchingNotification
+                                                         object:nil
+                                                          queue:nil usingBlock:^(NSNotification * _Nonnull note) {
+           
+           [SwiftAirshipPlugin takeOffWithLaunchOptions:note.userInfo];
```
rlepinski, line 15: Is this the right user info?

rlepinski (approved):
Could we bump versions of  the SDK to 16.1

ios/Classes/AirshipPlugin.m
oristanovic: yes

oristanovic:
> Could we bump versions of the SDK to 16.1

also bumped plugin version to 5.1

Ulrico972 (approved):

Issue #108, opened 2022-01-03
LateInitializationError: Local 'notification' has not been initialized.


spencer-sf:
# Preliminary Info

### What Airship dependencies are you using?
airship_flutter: ^4.0.0
pubspec.lock: airship_flutter: 4.4.0
(Know this isn't the latest version but looking at the repo this issue is still present in latest)

# Report

### What unexpected behavior are you seeing?
When receiving a push notification payload on the client after triggering a push from the Apptentive Message center using their airship integration, the deserialization of that payload fails because there is no `notification` object in the payload.

The instance member `notification` is non-nullable, but the `_fromJson` method uses a `late` declaration and a null check in conjunction when they shouldn't be if `notification` can be null as it is in my case.

<img width="610" alt="Screen Shot 2022-01-03 at 12 04 12 PM" src="https://user-images.githubusercontent.com/90425494/147958604-7fd3f0be-0bae-4638-b99e-ca6fbaee7ba8.png">

### What is the expected behavior?
Expect `notification` to be nullable if not present in payload

### What are the steps to reproduce the unexpected behavior?
Send a notification without a `notification` property

### Do you have logging for the issue?
<img width="1100" alt="Screen Shot 2022-01-03 at 11 47 35 AM" src="https://user-images.githubusercontent.com/90425494/147958003-915cebe0-ec53-4ba2-b7c6-31225f962190.png">

rlepinski:
Thanks for the report, we will take a look

Ulrico972:
Hello, we noticed the issue and will prepare a fix for this

Ulrico972:
Hello, we released the version 5.1.1 that fixes your issue.

PR #110, opened 2022-01-07
[GHI-108] make notification optional for pushReceivedEvent

Ulrico972:
### What do these changes do?
Makes notification optional for pushReceivedEvent

To fix a crash. Github Issue #108 

Build and run OK
Manual test to verify it's working fine.

### DIFF
--- a/lib/src/airship_flutter.dart
+++ b/lib/src/airship_flutter.dart
@@ -96,7 +96,7 @@ class NotificationResponseEvent {
 
 class PushReceivedEvent {
   final Map<String, dynamic>? payload;
-  final Notification notification;
+  final Notification? notification;
 
   const PushReceivedEvent._internal(this.payload, this.notification);
 


rlepinski (approved):

lib/src/airship_flutter.dart
```diff
@@ -96,7 +96,7 @@ class NotificationResponseEvent {
 
 class PushReceivedEvent {
   final Map<String, dynamic>? payload;
-  final Notification notification;
+  final Notification? notification;
```
spencer-sf, line 99: The `_fromJson` method will still throw an error with this as optional because the `late` initializer on line 106 requires notification to be set before return even if it's nullable

Ulrico972: That's right, thank you

m-skolnick (approved):

jyaganeh (approved):

PR #118, opened 2022-02-14
[MOBILE-2836] Preference center updates

oristanovic:
### What do these changes do?
Add a way to retrieve subscription lists and the preference center config

Feature compat with the native SDK

Tested with the test app to check that the data is correctly received

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -1,35 +1,35 @@
 package com.airship.flutter
 
-import android.graphics.Bitmap
-import android.webkit.WebView
+import android.annotation.SuppressLint
 import android.app.NotificationManager
 import android.content.Context
 import android.content.SharedPreferences
-import android.view.View
+import android.graphics.Bitmap
 import android.os.Build
+import android.view.View
+import android.webkit.WebView
 import androidx.core.app.NotificationManagerCompat
-import com.urbanairship.Autopilot
-import com.airship.flutter.events.ShowInboxMessageEvent
 import com.airship.flutter.events.ShowPreferenceCenterEvent
-import com.urbanairship.PendingResult
+import com.urbanairship.Autopilot
 import com.urbanairship.PrivacyManager
 import com.urbanairship.UAirship
 import com.urbanairship.analytics.CustomEvent
 import com.urbanairship.automation.InAppAutomation
+import com.urbanairship.channel.AttributeEditor
+import com.urbanairship.channel.TagGroupsEditor
+import com.urbanairship.contacts.Scope
 import com.urbanairship.json.JsonMap
 import com.urbanairship.json.JsonValue
-import com.urbanairship.channel.TagGroupsEditor
-import com.urbanairship.channel.AttributeEditor
-import com.urbanairship.contacts.Scope
 import com.urbanairship.messagecenter.MessageCenter
-import com.urbanairship.util.DateUtils
-import com.urbanairship.util.UAStringUtil
 import com.urbanairship.messagecenter.webkit.MessageWebView
 import com.urbanairship.messagecenter.webkit.MessageWebViewClient
 import com.urbanairship.preferencecenter.PreferenceCenter
-import com.urbanairship.preferencecenter.data.Item
-import com.urbanairship.preferencecenter.data.PreferenceCenterConfig
-import java.lang.NumberFormatException
+import com.urbanairship.reactive.Observable
+import com.urbanairship.reactive.Subscriber
+import com.urbanairship.util.DateUtils
+import com.urbanairship.util.UAStringUtil
+import io.flutter.embedding.engine.plugins.FlutterPlugin
+import io.flutter.plugin.common.BinaryMessenger
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
@@ -38,15 +38,8 @@ import io.flutter.plugin.common.PluginRegistry.Registrar
 import io.flutter.plugin.common.StandardMessageCodec
 import io.flutter.plugin.platform.PlatformView
 import io.flutter.plugin.platform.PlatformViewFactory
-import io.flutter.embedding.engine.plugins.FlutterPlugin
-import io.flutter.plugin.common.BinaryMessenger
 import io.flutter.plugin.platform.PlatformViewRegistry
 import kotlinx.coroutines.*
-import kotlinx.coroutines.flow.flow
-import kotlinx.coroutines.flow.toList
-import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 
 private const val TAG_OPERATION_GROUP_NAME = "group"
 private const val TAG_OPERATION_TYPE = "operationType"
@@ -142,7 +135,7 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
 
     private lateinit var context: Context
 
-    val scope: CoroutineScope = CoroutineScope(Dispatchers.Main)
+    val scope: CoroutineScope = CoroutineScope(Dispatchers.Main) + SupervisorJob()
 
     companion object {
         @JvmStatic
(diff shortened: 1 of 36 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -541,6 +545,101 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
...
+                it?.let {
+                    subscriptionLists["channel"] = it
+                }
+            }
+        }
+        if (subscriptionTypes.contains("contact")) {
+            val contactSubscriptionPendingResult = UAirship.shared().contact.getSubscriptionLists(true)
+            contactSubscriptionPendingResult.addResultCallback {
+                it?.let {
+                    subscriptionLists["contact"] = it
```
rlepinski, line 565: You should print on the other side to make sure the enum works properly

lib/src/airship_flutter.dart
```diff
@@ -351,5 +351,15 @@ class Airship {
   static Future<void> openPreferenceCenter(String preferenceCenterID) async {
     return await _channel.invokeMethod('openPreferenceCenter', preferenceCenterID);
   }
+
+  static Future<Map<String, dynamic>> getSubscriptionLists(List<String> subscriptionListTypes) async {
+    var lists = await (_channel.invokeMethod("getSubscriptionLists", subscriptionListTypes));
+    return Map<String, dynamic>.from(lists);
```
rlepinski, line 357: We should have a model object instead of using dynamic

lib/src/airship_flutter.dart
```diff
@@ -351,5 +351,15 @@ class Airship {
...
+  }
+
+  static Future<Map<String, dynamic>> getPreferenceCenterConfig(String preferenceCenterID) async {
+    var config = await _channel.invokeMethod('getPreferenceCenterConfig', preferenceCenterID);
+    return Map<String, dynamic>.from(config);
```
rlepinski, line 362: Same thing here, can we provide a model object?

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -621,6 +625,129 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin, RegistrationDelegate,
...
+                
+                if (item.itemType == .channelSubscription) {
+                    let subscriptionItem = item as! ChannelSubscriptionItem
+                    itemDict.updateValue(subscriptionItem.subscriptionID, forKey:"subscription_id")
+                } else if (item.itemType == .contactSubscription) {
+                    let subscriptionItem = item as! ContactSubscriptionItem
+                    itemDict.updateValue(subscriptionItem.subscriptionID, forKey:"subscription_id")
+                } else if (item.itemType == .contactSubscriptionGroup) {
+                    let subscriptionItem = item as! ContactSubscriptionGroupItem
+                    itemDict.updateValue(subscriptionItem.subscriptionID, forKey:"subscription_id")
```
rlepinski, line 708: Do we need a type?

rlepinski:
What about APIs to handle showing a preference center in flutter vs using the OOTB views?

oristanovic:
> What about APIs to handle showing a preference center in flutter vs using the OOTB views?

you mean something similar to this https://github.com/urbanairship/airship-flutter/blob/main/lib/src/inbox_message_view.dart ?

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -568,10 +587,135 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
...
 
+    private fun getSubscriptionLists(call: MethodCall, result: Result) {
+
+        val subscriptionTypes = call.arguments as List<String>
+
+        val subscriptionLists = mutableMapOf<String, Any>()
+
+        scope.launch {
+            val r1 = UAirship.shared().contact.getSubscriptionLists(true)
+            suspendCoroutine<Unit> { continuation ->
```
jyaganeh, line 610: I'm pretty sure this will end up making the requests and waiting for results in sequence, since the `suspendCoroutine`s will suspend this coroutine until the result callback is called. I think `async` seems like a better fit here, so that we can fetch in parallel. It should also be fine to just block on the `PendingResult`s instead of using result callbacks if we also move off of main dispatcher.

```kotlin
scope.launch(Dispatchers.Main) {
    val subscriptionLists = withContext(Dispatchers.IO) {
        val channelSubs = async {
            UAirship.shared().channel.getSubscriptionLists(true).get()
                ?.toList()
                ?: emptyList()
        }
        val contactSubs = async {
            UAirship.shared().contact.getSubscriptionLists(true).get()
                ?.mapValues { it.value.map(Scope::toString) }
                ?: emptyMap()
        }
       
       // Block until both requests complete and return a map
        mapOf(
            "channel" to channelSubs.await(),
            "contact" to contactSubs.await()
        )
    }

    // Callback on main dispatcher with result
    result.success(subscriptionLists)
}
```

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -568,10 +587,135 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
...
+    private fun setAutoLaunchDefaultPreferenceCenter(call: MethodCall, result: Result) {
+        val enabled = call.arguments as Boolean
+
+        val sharedPref: SharedPreferences = context.getSharedPreferences("shared_preferences", Context.MODE_PRIVATE)
+        sharedPref.edit().putBoolean(AUTO_LAUNCH_PREFERENCE_CENTER_KEY, enabled)
+
+        result.success(null)
+    }
+
+    private fun configDict(config: PreferenceCenterConfig): Map<String, Any> {
```
jyaganeh, line 656: This could probably be shortened a bit using some of Kotlin's fancy collection transformation APIs, but I'm not gonna block on it. Might be worth adding SDK methods that return raw pref center json to make this easier to wrangle in the future, though?

jyaganeh (approved):

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
rlepinski: 💯

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -566,12 +579,156 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
...
+            }
+        }
+        
+        dispatchGroup.notify(queue: .main, execute: {
+            result(subscriptionLists)
+        })
+    }
+    
+    private func editChannelSubscriptionLists(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        let subscriptionLists = call.arguments as! [[String:String]]
```
rlepinski, line 631: guard let with an error result?

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -566,12 +579,156 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
...
+        }
+    }
 
+    private func setAutoLaunchDefaultPreferenceCenter(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        guard let enabled = call.arguments as? Bool else {
+            result(nil)
+            return
+        }
+        
+        UserDefaults.standard.set(enabled, forKey: autoLaunchPreferenceCenterKey)
```
rlepinski, line 713: we should use our own suite for storage

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -135,6 +135,10 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
 
     private lateinit var context: Context
 
+    val sharedPreferences by lazy {
+        context.getSharedPreferences("shared_preferences", Context.MODE_PRIVATE)
```
rlepinski, line 139: this should be "com.urbanairship.flutter"

rlepinski:
The only other thing is a method to open a preference center

oristanovic:
> The only other thing is a method to open a preference center
@rlepinski this one?
https://github.com/urbanairship/airship-flutter/blob/a0935a868a0f7b18e9267d4522aa3481d646d445/lib/src/airship_flutter.dart#L417

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -593,7 +593,6 @@ class AirshipPlugin : MethodCallHandler, FlutterPlugin {
     private fun openPreferenceCenter(call: MethodCall, result: Result) {
         val preferenceCenterID = call.arguments as String
 
-        PreferenceCenter.shared().open(preferenceCenterID)
         PreferenceCenter.shared().openListener =  object : PreferenceCenter.OnOpenListener {
```
rlepinski, line 596: Needs to be set in Autopilot

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -580,6 +580,7 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin, PreferenceCenterOpenDe
             return
         }
         
+        PreferenceCenter.shared.openDelegate = self
```
rlepinski, line 583: needs to be set right after takeoff

rlepinski (approved):

Issue #115, opened 2022-02-09
Initialise Airship sdk at runtime with secret keys from remote source.


ailic88:
Is it possible to fetch "productionAppKey" "productionAppSecret" from remote source at runtime and do initialisation?
Or this data has to be part of "airshipconfig.properties" and "AirshipConfig.plist" files respectively for Android and iOS?

### What Airship dependencies are you using?
4.4.0

### What unexpected behavior are you seeing?
 There are no methods exposed for set configuration at runtime

rlepinski:
Its not possible right now in the plugin, but I will look into what it would take to make it possible. In our cordova plugin we make it possible to call takeOff from the cordova context, we then persist those config values. If takeOff is called again, the config wont take effect until next run. I assume that would work for you?

ailic88:
@rlepinski first thanks for quick response. When you say next run, do you mean on cold start("kill" app and run it again) or you mean on warm start(restart and re-initialise). For my app is not acceptable to keep keys locally, so if you could expose takeOff() method it would be perfect. Do you have any ETA when you can apply this change?

rlepinski:
Either, as long as the app has to start the lifecycle again (Android Application#onCreate, iOS didFinishLaunching). The underlying Airship SDK only supports being initialized once per app lifecycle at the moment. And once takeOff is called, we need to reinitialize the SDK during the beginning of the apps lifecycle in order to properly handle notifications. 

If you are just avoiding hardcoding credentials in your app, I don't see any problems with this approach. If you are trying to change what Airship app you point to at runtime, thats where things don't work out as nicely.

In terms of an ETA, I can't really provide that right now. I might have some time to try to get it to work this week but it's not trivial what needs to be done. Ill keep you updated.

ailic88:
@rlepinski this works for me. You can close this issue when new version with the changes is released. Any idea when it could happen?

rlepinski:
Working on wrapping up the release. Hopefully we will have something this week.

rlepinski:
Fixed in 5.3.0

PR #119, opened 2022-02-16
GH-115: Add support for takeOff directly from dart

rlepinski:
Adds a new `takeOff` method that you can pass the app key and secret into.  The credentials are combined with the config in the plist/properties file. Release/Debug credentials set in the config will take precedents, so in order to use this the app has to remove appkey/secret from the config before calling takeOff. Updates to the config after will not take until the next app init.

I tested on both iOS and Android using the sample App. I launched the app with normal config to see that it worked. I then removed the app key/secret and assigned it to a button click. I navigated the app to make sure we dont have any crashes, then I pressed the takeOff button to verify everything works. Next app launch everything worked before the takeOff button.

### DIFF
--- a/ios/Classes/AirshipEventHandler.swift
+++ b/ios/Classes/AirshipEventHandler.swift
@@ -1,0 +1,84 @@
+import Flutter
+import AirshipKit
+
+public class AirshipEventHandler: NSObject,
+                                  RegistrationDelegate,
+                                  DeepLinkDelegate,
+                                  PushNotificationDelegate,
+                                  MessageCenterDisplayDelegate {
+    func register() {
+        Airship.push.registrationDelegate = self
+        Airship.shared.deepLinkDelegate = self
+        Airship.push.pushNotificationDelegate = self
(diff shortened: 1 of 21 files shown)


ios/Classes/AirshipEventHandler.swift
```diff
@@ -0,0 +1,84 @@
...
+    public func displayMessageCenter(forMessageID messageID: String, animated: Bool) {
+        self.showMessage(forID: messageID)
+    }
+
+    public func displayMessageCenter(animated: Bool) {
+        self.showInbox()
+    }
+
+    public func dismissMessageCenter(animated: Bool) {
+        //
```
jyaganeh, line 63: Should this be doing something?

jyaganeh (approved):

Ulrico972 (approved):

ios/Classes/AirshipEventHandler.swift
rlepinski: Maybe, its how it works now

PR #123, opened 2022-03-28
[MOBILE-2196] Support handling background notifications 

jyaganeh:
### What do these changes do?

Adds support for handling messages received while the app is not in the foreground by:
* Using `AirshipBackgroundExecutor`, which handles executing dart callbacks in a background Isolate. This is started in `FlutterAutopilot.onAirshipReady` if autopilot is started up while the app is backgrounded.
* Intercept pushes via overriding `onCreateNotification` on the notification provider registered in `FlutterAutopilot`. 
    * If foregrounded or no background handler is registered, we use default message handling provided by `AirshipNotificationProvider`.
    * Otherwise, we execute the registered background message handler callback on the background Isolate.
* A new `Airship.setBackgroundMessageHandler(handler)` method that allows developers to set a background message handler (and also starts up the background isolate). The user can inspect the message data and decide whether they want to handle the push or fall back to default handling, but any logic in their background message handler needs to return quickly to avoid ANRs from the service.

The Airship plugin is responsible for telling the background Isolate to start up and providing it with an isolate callback so that the flutter side can be notified when the isolate is ready for processing, and a message callback, which points to the user's top-level background message handler.

To support background push handling.

Ran sample app on Android and sent pushes in the foreground, background, and after rebooting (without relaunching the app manually). Also ran on iOS to make sure this didn't unexpectedly change anything with foreground or background push.


android/src/main/kotlin/com/airship/flutter/FlutterAutopilot.kt
```diff
@@ -12,34 +12,59 @@ import com.urbanairship.push.*
...
             EventManager.shared.notifyEvent(InboxUpdatedEvent())
         }
 
-        airship.pushManager.notificationProvider = object : AirshipNotificationProvider(UAirship.getApplicationContext(), airship.airshipConfigOptions) {
+        airship.pushManager.notificationProvider = object : AirshipNotificationProvider(appContext, airship.airshipConfigOptions) {
+            override fun onCreateNotification(
+                context: Context,
+                arguments: NotificationArguments
+            ): NotificationResult {
+                if (appContext.isAppInForeground()) {
```
rlepinski, line 56: I think generally people will want the opposite of what we providing here. In the foreground they might want to do something special, but if they are in the background show the notification.

I am worried we are over complicating this and should just tell them we received a push in the background to do custom processing.

rlepinski (approved):

Ulrico972 (approved):

PR #128, opened 2022-04-26
Add custom preference center

khmMouna:
### What do these changes do?
Add custom preference center screen

### DIFF
--- a/example/lib/screens/preference_center.dart
+++ b/example/lib/screens/preference_center.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:airship_flutter/airship_flutter.dart';
 import 'package:airship_example/styles.dart';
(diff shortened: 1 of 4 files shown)


example/lib/screens/preference_center.dart
```diff
@@ -0,0 +1,250 @@
...
+    ScopedSubscriptionListEditor editor = Airship.editContactSubscriptionLists();
+    if (isSubscribed) {
+      editor.subscribe(subscriptionId, scopes);
+      activeContactSubscriptions.putIfAbsent(subscriptionId, () => scopes);
+    } else {
+      editor.unsubscribe(subscriptionId, scopes);
+      activeContactSubscriptions.remove(subscriptionId);
+    }
+    editor.apply();
+    //setState(() {});
```
jyaganeh, line 90: should this be uncommented or removed?

example/lib/screens/preference_center.dart
```diff
@@ -0,0 +1,250 @@
...
+    return ListTile(
+        title: Text('${item.display.title}', style:TextStyle(fontWeight: FontWeight.bold)),
+        subtitle: Text('${item.display.subtitle}'),
+    );
+  }
+
+  Widget item(IndexPath indexPath) {
+    List<PreferenceCenterItem> items = preferenceCenterConfig.sections[indexPath.section].items;
+    PreferenceCenterItem item = items[indexPath.item];
+    switch (item.type) {
```
jyaganeh, line 171: CI is complaining about a missing return for this method if we don't match any of the cases. Might need to throw or return a null that we can ignore higher up?

jyaganeh (approved):
Nice! The sample seems easy to understand and tweak 👍

example/lib/screens/preference_center.dart
khmMouna: Yes 👍  it should be uncommented. I did some other improvements if you want to take a look

jyaganeh (approved):
👍

PR #133, opened 2022-06-25
Mobile 3122 expand takeoff on flutter

shadracknx:
### What do these changes do?

- Adds protobuf as dependence
- Parse config using airship takeOff

Adds a consistent takeOff API to flutter

Runs with the sample app


shadracknx:
Current changes are mainly for android. This so that , all feedback that can be applied to iOS from android does not need duplicate review ahead of time.

lib/src/airship_flutter.dart
```diff
@@ -295,11 +296,31 @@ class Airship {
...
   /// Returns true if Airship has been initialized, otherwise returns false.
-  static Future<bool> takeOff(String appKey, String appSecret) async {
-    Map<String, String> args = {"app_key": appKey, "app_secret": appSecret};
+  /// For backwards compatibility [keyConfig] can be appKey[String] or [AirshipConfig]
+  /// in next major release airship will migrate to [AirshipConfig]
+  static Future<bool> takeOff(final Object keyConfig,
+      [final String appSecret = ""]) async {
+    assert(
+      keyConfig is String || keyConfig is AirshipConfig,
+      "For backwards compatibility [appKey] can be [String] or [AirshipConfig]",
```
jyaganeh: This could be `Airship failed to takeOff! An appKey String or valid AirshipConfig are required.`?

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -239,36 +266,47 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
-        println(config.values);
-        val appKey = config["app_key"] as String
-        val appSecret = config["app_secret"] as String
+        val configByteArray = call.arguments as ByteArray
+        // todo add serialization
+        // println(config.values);
+        // val appKey = config["appKey"] as String
+        // val appSecret = config["appSecret"] as String
+
+        /// 1.Get AirshipEnv config
```
jyaganeh, line 300: Are these still TODOs, or can we remove these commented lines?

example/lib/config/airship.config.sample.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:airship_flutter/airship_flutter.dart';
+
+abstract class Config {
+  static AirshipConfig airship = AirshipConfig(
+      defaultEnv: AirshipEnv(
+        /// Add your appKey
```
jyaganeh, line 6: nitpick: this line should be indented to match the next line

ios/Classes/PluginConfig.swift
```diff
@@ -1,30 +1,175 @@
...
     private static func read<T>(_ key: Keys) -> T? {
         return defaults.object(forKey: key.rawValue) as? T
     }
-    
+
     private static func write(_ key: Keys, value: Any?) {
-        defaults.set(value, forKey: key.rawValue)
+        if let value = value {
+            defaults.set(value, forKey: key.rawValue)
+                } else {
```
jyaganeh, line 40: nitpick: else is indented too far

ios/Classes/PluginConfig.swift
```diff
@@ -1,30 +1,175 @@
...
+                    defaults.removeObject(forKey: key.rawValue)
+                }
+    }
+}
+
+/* Copyright Airship and Contributors */
```
jyaganeh, line 46: nitpick: move to the top of the file

lib/src/airship_flutter.dart
```diff
@@ -295,12 +293,20 @@ class Airship {
...
+  /// Initializes Airship with an [AirshipConfig] or
+  /// [appKey] and [appSecret].
   ///
   /// Returns true if Airship has been initialized, otherwise returns false.
-  static Future<bool> takeOff(String appKey, String appSecret) async {
-    Map<String, String> args = {"app_key": appKey, "app_secret": appSecret};
-    return await _channel.invokeMethod('takeOff', args);
+  /// For backwards compatibility [config] can be appKey[String] or [AirshipConfig]
+  /// in next major release airship will migrate to [AirshipConfig]
+  static Future<bool> takeOff(final AirshipConfig config) async {
```
jyaganeh, line 302: do we need another method that accepts `String appKey, String appSecret`?

android/src/main/kotlin/com/airship/flutter/ConfigParser.kt
```diff
@@ -0,0 +1,190 @@
...
+        }
+        val channelId = notificationConfig.opt("defaultChannelId").string
+        if (channelId != null) {
+            builder.setNotificationChannel(channelId)
+        }
+    }
+
+    @ColorInt
+    fun getHexColor(hexColor: Int?, @ColorInt defaultColor: Int): Int {
+        return hexColor?:defaultColor
```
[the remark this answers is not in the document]
shadracknx: Will format

lib/src/airship_flutter.dart
shadracknx: No, this needs to be removed

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -354,7 +394,10 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
         result.success(null)
     }
 
-    private fun applyTagGroupOperations(editor: TagGroupsEditor, operations: ArrayList<Map<String, Any?>>) {
+    private fun applyTagGroupOperations(
```
rlepinski, line 397: It would be great to just do a commit with lint stuff applied so we can stop having it distract from the actual changes

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -27,41 +41,178 @@ class ConfigManager(private val context: Context) {
...
+                    pluginConfig.production.let {
+                        this.setProductionAppKey(it.appKey)
+                            .setProductionAppSecret(it.appSecret)
+                            .setProductionLogLevel(it.logLevel.value)
+                            .build()
+                    }
+
+
+                    this.setSite(
+                        when (pluginConfig.site!!) {
```
rlepinski, line 107: can we mark this as nonnull somewhere else? maybe fallback to US if not set? we usually avoid force unwrapping when possible

android/build.gradle
```diff
@@ -59,4 +59,6 @@ dependencies {
     implementation "com.urbanairship.android:urbanairship-automation:$airship_version"
     implementation "com.urbanairship.android:urbanairship-preference-center:$airship_version"
     implementation "androidx.datastore:datastore-preferences:1.0.0"
+    implementation "androidx.datastore:datastore:1.0.0"
```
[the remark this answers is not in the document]
shadracknx: Okay, I only kept around to use with to be able to use already saved credentials

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
shadracknx: true

shadracknx:
I have tested flutter takeOff everything works as expected.

I am bit confused on two exceptions I am getting when I cold boot the emulator on android. 

1. `com.urbanairship.http.RequestException: Unable to perform request: missing URL`  and 
2. `Unhandled Exception: PlatformException(AIRSHIP_GROUNDED, TakeOff not called., null, null)` only in message codec.

Am not sure I know what is going own, debugging

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,42 +28,174 @@ class ConfigManager(private val context: Context) {
...
+                    }
+
+                    pluginConfig.urlAllowListScopeOpenUrlList.let {
+
+                        this.setUrlAllowListScopeOpenUrl(it.toTypedArray())
+                    }
+
+                    this.setEnabledFeatures(pluginConfig.featuresEnabledList.supportValue())
+
+                    this.setAppStoreUri(Uri.parse(pluginConfig.android.appStoreUri))
```
jyaganeh, line 114: It'd be good to wrap the `Uri.parse` in a try/catch here, so we can log a helpful error message. It might be worth re-throwing so that it's obvious that something is broken, though.

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,42 +28,174 @@ class ConfigManager(private val context: Context) {
...
+                    pluginConfig.urlAllowlistScopeJavascriptInterfaceList.let {
+                        this.setUrlAllowListScopeJavaScriptInterface(it.toTypedArray())
+                    }
+
+                    pluginConfig.urlAllowListScopeOpenUrlList.let {
+
+                        this.setUrlAllowListScopeOpenUrl(it.toTypedArray())
+                    }
+
+                    this.setEnabledFeatures(pluginConfig.featuresEnabledList.supportValue())
```
jyaganeh, line 112: I think we may want something like this, unless the default of `FEATURE_ALL` is being handled by the Proto definition:

``` kotlin
if (pluginConfig.featuresEnabledCount > 0) {
    config.setEnabledFeatures(pluginConfig.featuresEnabledList.supportValue())
} else {
    config.setEnabledFeatures(PrivacyManager.FEATURE_ALL)
}
```

ios/Classes/PluginConfig.swift
```diff
@@ -1,30 +1,174 @@
...
+    }
+}
+
+
+
+extension Config {
+    static func parse(_ configDict: AirshipConfig) throws -> Config {
+        let airshipConfig = Config.default()
+        
+        airshipConfig.parseDefaultEnv(configDict.defaultEnv)
```
jyaganeh, line 53: Not sure if Swift protobuf is similar to Kotlin in terms of empty/default values vs nullable fields, but do we need to handle those here, too?

example/lib/config/airship.config.sample.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:airship_flutter/airship_flutter.dart';
+
+abstract class Config {
+  static AirshipConfig airship = AirshipConfig(
+      defaultEnv: AirshipEnv(
```
jyaganeh, line 5: I'd maybe move `inProduction` to the top, so that it's harder to overlook?

jyaganeh (commented):
Checked out your branch, poked around at Android, and left some more comments.
I think we'll need to tweak Android (and maybe iOS) to make sure that we're creating an `AirshipConfig` from the options that were set in the `dart` config, without any default/empty values that might be coming from Protobuf.

If you print the parsed dart config and the generated `AirshipConfig`, it should be easy to spot if anything is getting set in the `AirshipConfig` that might be unexpected. It'd be good to verify the same on iOS, as well. I dropped a Kotlin extension below that adds a debug string method for `AirshipConfig`.

<details><summary>AirshipConfig Extension</summary>

```kotlin
fun AirshipConfigOptions.debugString(): String {
    return "AirshipConfigOptions{" +
        "appKey='$appKey',\n" +
        "appSecret='$appSecret',\n" +
        "deviceUrl='$deviceUrl',\n" +
        "analyticsUrl='$analyticsUrl',\n" +
        "remoteDataUrl='$remoteDataUrl',\n" +
        "walletUrl='$walletUrl',\n" +
        "chatUrl='$chatUrl',\n" +
        "chatSocketUrl='$chatSocketUrl',\n" +
        "appStoreUri=$appStoreUri,\n" +
        "allowedTransports=$allowedTransports,\n" +
        "customPushProvider=$customPushProvider,\n" +
        "urlAllowList=$urlAllowList,\n" +
        "urlAllowListScopeJavaScriptInterface=$urlAllowListScopeJavaScriptInterface,\n" +
        "urlAllowListScopeOpenUrl=$urlAllowListScopeOpenUrl,\n" +
        "analyticsEnabled=$analyticsEnabled,\n" +
        "backgroundReportingIntervalMS=$backgroundReportingIntervalMS,\n" +
        "logLevel=$logLevel,\n" +
        "autoLaunchApplication=$autoLaunchApplication,\n" +
        "channelCreationDelayEnabled=$channelCreationDelayEnabled,\n" +
        "channelCaptureEnabled=$channelCaptureEnabled,\n" +
        "dataCollectionOptInEnabled=$dataCollectionOptInEnabled,\n" +
        "enabledFeatures=$enabledFeatures,\n" +
        "extendedBroadcastsEnabled=$extendedBroadcastsEnabled,\n" +
        "notificationIcon=$notificationIcon,\n" +
        "notificationLargeIcon=$notificationLargeIcon,\n" +
        "notificationAccentColor=$notificationAccentColor,\n" +
        "notificationChannel='$notificationChannel',\n" +
        "inProduction=$inProduction,\n" +
        "requireInitialRemoteConfigEnabled=$requireInitialRemoteConfigEnabled,\n" +
        "fcmFirebaseAppName='$fcmFirebaseAppName'" +
        "}";
}
```

</details>

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,42 +28,174 @@ class ConfigManager(private val context: Context) {
...
-            val appCredentialsOverride =
-                context.airshipFlutterPluginStore.data.map { preferences ->
-                    Pair(preferences[APP_KEY], preferences[APP_SECRET])
-                }.first()
+            /// Plugin config stored/ from dart config
+            val pluginConfig = context.flutterPluginStore.data.first()
 
             val config = AirshipConfigOptions.newBuilder()
-                .applyDefaultProperties(context)
+                //.applyDefaultProperties(context)
```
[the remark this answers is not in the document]
shadracknx: I thought all we needed was configuration from Dart for both IOS and Android. 

@rlepinski do we still need to read from `assets/airshipconfig.properties` for backwards compatibility?

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,42 +28,174 @@ class ConfigManager(private val context: Context) {
...
+                .setRequireInitialRemoteConfigEnabled(true)
```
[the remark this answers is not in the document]
shadracknx: defaults the config property from dart Airship config `setRequireInitialRemoteConfigEnabled`

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
shadracknx: The only thing `Uri.parse` throws is a `NullPointerException` – if uriString is null 
I think just checking that would be better.

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
shadracknx: It is done through protobuf

ios/Classes/PluginConfig.swift
shadracknx: Applying all android changes into IOS

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,49 +31,246 @@ class ConfigManager(private val context: Context) {
...
+                    pluginConfig.inProduction.let {
+                        this.setInProduction(it)
+                    }
+
+                    pluginConfig.development.let {
+                        if (it.isNotEmptyOrPartial) {
+                            this.setDevelopmentAppKey(it.appKey)
+                                .setDevelopmentAppSecret(it.appSecret)
+                                .setDevelopmentLogLevel(it.logLevel.value)
+                                .build()
```
jyaganeh, line 64: `build()` doesn't need to be called here and is throwing warnings into the logs about the allow lists not being set on the config

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,49 +31,246 @@ class ConfigManager(private val context: Context) {
...
+                        }
+                    }
+
+                    pluginConfig.production.let {
+                        if (it.isNotEmptyOrPartial) {
+                            this.setProductionAppKey(it.appKey)
+                                .setProductionAppSecret(it.appSecret)
+                                .setProductionLogLevel(it.logLevel.value)
+                                .build()
```
jyaganeh, line 73: same here

android/src/main/kotlin/com/airship/flutter/ConfigManager.kt
```diff
@@ -26,49 +31,246 @@ class ConfigManager(private val context: Context) {
...
+                    }
+
+                    /// since proto buf defaults to empty string,
+                    // check for empty string instead of null
+                    pluginConfig.defaultEnv.let {
+                        if (it.isNotEmptyOrPartial) {
+                            this.setAppKey(it.appKey)
+                                .setAppSecret(it.appSecret)
+                                .setLogLevel(it.logLevel.value)
+                                .build()
```
jyaganeh, line 84: and here

ios/Classes/AirshipAutopilot.swift
```diff
@@ -3,39 +3,45 @@ import AirshipKit
...
 public class AirshipAutopilot: NSObject {
+
     
     @objc
     public static var launchOptions: [UIApplication.LaunchOptionsKey : Any]?
-
+    
     @objc
     public static func attemptTakeOff() {
+        print("Taking off -----------------------------")
```
jyaganeh, line 13: this debug log should be removed

ios/Classes/AirshipAutopilot.swift
```diff
@@ -3,39 +3,45 @@ import AirshipKit
...
-        config.productionAppSecret = config.productionAppSecret ?? PluginConfig.appSecret
-        
-        guard config.validate() else {
+        guard let configDict = PluginStore.protoConfig else {
             return
         }
+        do {
+            
+            let config = try Config.parse(configDict)
+            AirshipLogger.debug("Taking off! \(config)")
```
jyaganeh, line 24: don't need to log the whole config here

ios/Classes/PluginConfig.swift
```diff
@@ -1,30 +1,191 @@
+/* Copyright Airship and Contributors */
 import Foundation
 import AirshipKit
+import SwiftProtobuf
+
 
 class PluginConfig {
```
jyaganeh, line 7: It looks like this class isn't being used any more, unless I'm just missing where it's used. Do we still need it?

ios/Classes/PluginStore.swift
```diff
@@ -0,0 +1,69 @@
+/* Copyright Airship and Contributors */
+
+import Foundation
+import AirshipKit
+
+class PluginStore {
+    
+    private static let defaults = SwiftAirshipPlugin.defaults
+    
+    public static var config: [String: Any]? {
```
jyaganeh, line 10: Not seeing this being used anywhere, either. Maybe we can get rid of it, too?

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -164,94 +164,101 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin, PreferenceCenterOpenDe
...
+        AirshipLogger.debug("\(String(describing:call.arguments))")
+        guard let configDict = call.arguments as? FlutterStandardTypedData else {
             result(false)
-            return
+            fatalError("Invalid arguments: \(String(describing: call.arguments))")
+        }
+        do{
+            PluginStore.protoConfig = try AirshipConfig.init(serializedData:configDict.data)
+            
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.016) {  AirshipAutopilot.attemptTakeOff()
```
jyaganeh, line 177: Why did we need a delay here?

jyaganeh (commented):
looking pretty good! just had a few more comments/questions

ios/Classes/PluginConfig.swift
shadracknx: True

ios/Classes/PluginStore.swift
shadracknx: Yeah, that is a leftover from config file setup

ios/Classes/SwiftAirshipPlugin.swift
shadracknx: Actually was trying to test, something I noticed, cold start is not always registering the channel on the first run. Try run it from scratch and let me know whether it is consistent every time.

lib/src/airship_flutter.dart
```diff
@@ -586,7 +580,7 @@ class Airship {
   }
 
   /// Returns a [List] with the enabled features.
-  static Future<List<String>> getEnabledFeatures() async {
+  static FutureOr<List<String>> getEnabledFeatures() async {
```
rlepinski, line 583: what is this change for?

ios/Classes/PluginConfig.swift
```diff
@@ -1,30 +1,190 @@
...
+    private func parseDefaultEnv(_ defaultEnv: AirshipEnv){
+        if(!defaultEnv.isEmptyOrPatial){
+        defaultAppKey = defaultEnv.appKey
+        defaultAppSecret = defaultEnv.appSecret
+        productionLogLevel = defaultEnv.logLevel.value()
+        }
+    }
+    
+    private func parseDevelopmentEnv(_ developmentEnv: AirshipEnv){
+        if(!developmentEnv.isEmptyOrPatial){
```
rlepinski, line 88: formatting is messed up here

ios/Classes/PluginStore.swift
```diff
@@ -0,0 +1,60 @@
...
+    private static let defaults = SwiftAirshipPlugin.defaults
+    
+    public static var protoConfig:  AirshipConfig? {
+        get {
+            do{
+                let data = defaults.value(forKey: "protoConfig")
+                guard data != nil else {
+                    return nil
+                }
+                return try AirshipConfig.init(serializedData: data as! Data);
```
rlepinski, line 17: guard let data = defaults.value(forKey: "protoConfig") as? Data else {
  return nil
}
AirshipConfig(serializedData: data)

lib/src/airship_flutter.dart
shadracknx: Allows return type to be either `Future<List<String>>` or `List<String>` while keeping type safety.

That is inline with what I saw, occasionally it would through a type related error that solved it.

rlepinski (approved):
With merge into next

PR #150, opened 2022-12-06
Annotate background message handler with @pragma('vm:entry-point')

cs-nuuk:
### What do these changes do?
With Flutter 3.3 you have to annotate methods that will be used from native code with `@pragma('vm:entry-point')`. I've added this annotation to `_backgroundMessageIsolateCallback()`.

When you use this plugin with Flutter 3.3 you will get the following error message with release builds.
```
E/flutter (27564): [ERROR:flutter/shell/common/shell.cc(93)] Dart Error: Dart_LookupLibrary:library 'package:airship_flutter/src/airship_flutter.dart' not found.
```
This happens because the compiler does not know that you call this function from native code so it removes it. 

Build the example app as release with Flutter 3.3 and view the logs. The error message will be gone.

- https://github.com/dart-lang/sdk/blob/master/runtime/docs/compiler/aot/entry_point_pragma.md
- https://github.com/firebase/flutterfire/issues/9446#issuecomment-1240554285

### DIFF
--- a/lib/src/airship_flutter.dart
+++ b/lib/src/airship_flutter.dart
@@ -211,6 +211,7 @@ class PushReceivedEvent {
   }
 }
 
+@pragma('vm:entry-point')
 void _backgroundMessageIsolateCallback() {
   WidgetsFlutterBinding.ensureInitialized();
 


rlepinski:
Thanks!

I assume its fine, but if we merge this does it now require using flutter 3.3+ or is it still backwards compatible with 3.0?

oristanovic:
@cs-nuuk Could you please fill in this contribution agreement? https://docs.google.com/forms/d/e/1FAIpQLScErfiz-fXSPpVZ9r8Di2Tr2xDFxt5MgzUel0__9vqUgvko7Q/viewform

oristanovic (approved):

PR #166, opened 2023-04-18
[MOBILE-3551] Remove the contact subscription to fix the switch state

Apekka:
Following a customer issue where the Preference center at Contact level does not work properly

### What do these changes do?
Before, when unsubscribing to a list, at a contact level, it only put an empty scope list as value, but the subscriptionId was still in the list.
Since the switch button was checking if the subscriptionId is in the list, the switch stayed enabled

So I made a change to remove the subscriptionId if you unsubscribe to it.

I also added a "app" scope by default, because the SDK does not subscribe if there is no scope in the Contact subscriptions.

To fix the Preference Center in Flutter

By running the sample app, and using the "preference_center" id, that is contact level

### DIFF
--- a/example/lib/screens/preference_center.dart
+++ b/example/lib/screens/preference_center.dart
@@ -129,9 +129,8 @@ class _PreferenceCenterState extends State<PreferenceCenter>
       title: Text('${item.display.title}',
           style: TextStyle(fontWeight: FontWeight.bold)),
       subtitle: Text('${item.display.subtitle}'),
-      value: isSubscribedContactSubscription(item.subscriptionId, []),
+      value: isSubscribedContactSubscription(item.subscriptionId, ["app"]),
       onChanged: (bool value) {
-        //TODO keep looking in the scopes list
         onPreferenceContactSubscriptionItemToggled(
             item.subscriptionId, ["app"], value);
       },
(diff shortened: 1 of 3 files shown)


example/lib/screens/preference_center.dart
```diff
@@ -130,10 +129,10 @@ class _PreferenceCenterState extends State<PreferenceCenter>
       title: Text('${item.display.title}',
           style: TextStyle(fontWeight: FontWeight.bold)),
       subtitle: Text('${item.display.subtitle}'),
-      value: isSubscribedContactSubscription(item.subscriptionId, []),
+      value: isSubscribedContactSubscription(item.subscriptionId, ["app"]),
```
jyaganeh, line 132: I think the scopes array here should be read from the `item`, but it looks like we may have missed the `scopes` field?

Here's where that's defined in the contact subscription items on [Android](https://github.com/urbanairship/android-library-dev/blob/main/urbanairship-preference-center/src/main/java/com/urbanairship/preferencecenter/data/Item.kt#L161) and [iOS](https://github.com/urbanairship/ios-library-dev/blob/main/Airship/AirshipPreferenceCenter/Source/data/Items.swift#L221).

It looks like `isSubscribedContactSubscription` will do the right thing if we pass it the item scopes.

jyaganeh (changes requested):

rlepinski (approved):

PR #172, opened 2023-05-19
Patch release 6.3.2

Apekka:
### What do these changes do?
Bump the minimal version of intl to >=0.18.0 make it compatible SDK 3.10.0

So a customer can use Flutter SDK 3.10.0

By running the sample app and testing it

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  intl: ">=0.18.0"
+  intl: "^0.18.0"
 
 dev_dependencies:
   flutter_test:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -12,7 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  intl: "^0.18.0"
+  intl: '>=0.15.7 <1.0.0'
 
 dev_dependencies:
   flutter_test:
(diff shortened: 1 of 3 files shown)


rlepinski (approved):

rlepinski:
@andrewescutia good point, what about just newer than 0.15.7?

pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  intl: ">=0.15.7 <0.18.0"
+  intl: "^0.18.0"
```
rlepinski, line 15: Lets do `>=0.15.7 <1.0.0` and hope they follow proper semantic versioning

rlepinski: I think we can remove this in the next release if we just expect a locale tag instead of a locale object

andrewescutia: According to their versioning doc:

Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API. For simplicity’s sake, avoid using + after the version reaches 1.0.0.

https://dart.dev/tools/pub/versioning#semantic-versions

so with that reasoning I think >=0.15.7 <0.19.0 would be the safest bet since 0.20 could be considered a breaking change deployment

rlepinski:
Will update to `>=0.15.7 <1.0.0` then try to remove the dependency in our bigger release

jyaganeh (approved):

rlepinski (approved):

rlepinski (approved):

PR #174, opened 2023-08-11
[MOBILE-3858] Add support for feature flag

Ulrico972:
### What do these changes do?
Add support for feature flag

To support feature flag

Manual testing


ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -405,6 +405,12 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
             ) as? AirshipJSON
             return result?.unWrap()
 
+        // Feature Flag
+        case "featureFlag#flag":
+            return try await AirshipProxy.shared.featureFlagManager.flag(
+                name: try call.requireStringArg()
+            )
```
rlepinski, line 412: I think we need to jsonify this result. `JSONValue.wrapOpt(result).unwrap` or whatever

lib/src/feature_flag.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  /// Indicates whether the flag exists in the current flag listing or not.
+  final bool exists;
+
+  /// Optional variables associated with the flag.
+  final Map<String, dynamic>? variables;
+
+  const FeatureFlag._internal(this.isEligible, this.exists, this.variables);
+
+  static FeatureFlag fromJson(Map<String, dynamic> json) {
```
rlepinski, line 16: pass in dynamic here

lib/src/feature_flag.dart
```diff
@@ -0,0 +1,37 @@
...
+    if (json == null) {
+      return null;
+    }
+
+    try {
+      var isEligible = json["isEligible"];
+      var exists = json["exists"];
+      var variables = Map<String, dynamic>.from(json["variables"]);
+      return FeatureFlag._internal(isEligible, exists, variables);
+    } catch (e, s) {
```
rlepinski, line 26: I think we should just let it crash?

lib/src/feature_flag.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  /// Indicates whether the flag exists in the current flag listing or not.
+  final bool exists;
+
+  /// Optional variables associated with the flag.
+  final Map<String, dynamic>? variables;
+
+  const FeatureFlag._internal(this.isEligible, this.exists, this.variables);
+
+  static FeatureFlag? fromJson(Map<String, dynamic>? json) {
```
rlepinski, line 16: I would make this ` static FeatureFlag fromJson(dynamic json)`

rlepinski:
I updated it with my feedback since ill be out. Here is the usage:
```
Airship.featureFlagManager.flag("another_rad_flag").then((value) => {
    debugPrint("flag: $value")
}).catchError((error) => {
    debugPrint("flag error: $error")
});
```

Ulrico972:
We have an issue on iOS, the proxy has a typo in one feature flag key. I think ryan found it but I'm not sure how he planed to handle it.
The best way would be to fix the proxy but it might be used already in other frameworks. So I'm checking them to see what's best. Maybe I'll find the workaround in one of them.

jyaganeh (approved):

PR #177, opened 2023-08-30
[MOBILE-3854] Add editTags method

Ulrico972:
### What do these changes do?
Add editTags method

To keep updated

I tested this with a local version of the proxy on android and iOS

### DIFF
--- a/lib/src/airship_channel.dart
+++ b/lib/src/airship_channel.dart
@@ -35,11 +35,13 @@ class AirshipChannel {
   }
 
   /// Adds channel tags.
+  /// Deprecated. Use editTags() instead.
   Future<void> addTags(List<String> tags) async {
     return await _module.channel.invokeMethod('channel#addTags', tags);
   }
 
   /// Removes channel tags.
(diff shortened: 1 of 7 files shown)


lib/src/airship_channel.dart
```diff
@@ -34,15 +35,22 @@ class AirshipChannel {
...
   }
 
   /// Removes channel tags.
+  /// Deprecated. Use editTags() instead.
   Future<void> removeTags(List<String> tags) async {
     return await _module.channel.invokeMethod('channel#removeTags', tags);
   }
 
+  /// Creates a [TagEditor] to modify the device tags.
+  TagEditor editTags() {
```
ilitvinenko, line 50: do we want to provide an API with closure that automatically call apply in the end? it was very handy to use in native sdks

Ulrico972: In my opinion, I think we want to keep the same structure than our other editors for this one, but I might be an idea for a next version

Apekka (approved):

Issue #186, opened 2023-10-02
Airship.messageCenter.onDisplay not parsing DisplayMessageCenterEvent properly


tomalabaster:
# Preliminary Info

### What Airship dependencies are you using?
airship_flutter: 7.1.1

### What are the versions of any relevant development tools you are using?
VS Code
Flutter 3.10.6
iOS 16.4 Simulator
Sending push with Message Center message in a TEST project to test group using the "Send test" button

# Report

### What unexpected behavior are you seeing?
When tapping on a push notification sent with a Message Center message, we get the following exception:

```
type '_Map<Object?, Object?>' is not a subtype of type 'Map<String, dynamic>'

#0      AirshipMessageCenter.onDisplay.<anonymous closure>
airship_message_center.dart:70
#1      _MapStream._handleData (dart:async/stream_pipe.dart:213:31)
#2      _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:153:13)
#3      _rootRunUnary (dart:async/zone.dart:1407:47)
#4      _CustomZone.runUnary (dart:async/zone.dart:1308:19)
#5      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1217:7)
#6      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
#7      _DelayedData.perform (dart:async/stream_impl.dart:515:14)
#8      _PendingEvents.handleNext (dart:async/stream_impl.dart:620:11)
#9      _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:591:7)
#10     _rootRun (dart:async/zone.dart:1391:47)
#11     _CustomZone.run (dart:async/zone.dart:1301:19)
#12     _CustomZone.runGuarded (dart:async/zone.dart:1209:7)
#13     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:asy<…>
```

Our code listening to this stream looks like:

```dart
Airship.messageCenter.onDisplay.listen((event) {
  if (event.messageId != null) {
    // navigate to our message enter
  }
});
```

### What is the expected behavior?
Events go through as expected

### What are the steps to reproduce the unexpected behavior?
Send a push notification from the Airship dashboard with an associated Message Center message

### Do you have logging for the issue?
See above

Apekka:
Hello, we fixed the issue and will do a patch release asap, I'll let you know when the release is up

Apekka:
[...]

PR #187, opened 2023-10-04
Fix event exception

Apekka:
Following an issue where the event stream wasn't properly parsed : https://github.com/urbanairship/airship-flutter/issues/186

Tested it with a push and message center action, and works properly now.


jyaganeh (approved):

rlepinski:
The constructor should take a dynamic instead and handle parsing within that. That will match the other events.

rlepinski (approved):

PR #190, opened 2023-11-03
Release 7.2.0

rlepinski:
Still need to verify the changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Flutter Plugin Changelog
 
+## Version 7.1.3 - November 6, 2023
+Patch release to fix a crash on iOS when using FlutterEngineGroups.
+
+### Changes
+- Fixed crash when using FlutterEngineGroups
+
 ## Version 7.1.2 - October 6, 2023
 Patch release that fixes an exception occuring while listening to the `Airship.messageCenter.onDisplay` event
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 Minor release that improves Feature Flag support on iOS.
 
 ### Changes
-- Updated Airship iOS SDK to 17.6.1.
+- Updated iOS SDK to 17.6.1.
+- Updated Android SDK to 17.5.0
 - Added `Airship.featureFlagManager.trackInteraction(flag)` method to track interaction events
 - Fixed crash when using FlutterEngineGroups
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Flutter Plugin Changelog
 
-## Version 7.1.3 - November 6, 2023
-Patch release to fix a crash on iOS when using FlutterEngineGroups.
+## Version 7.2.0 - November 27, 2023
+Minor release that improves Feature Flag support on iOS.
 
 ### Changes
+- Updated Airship iOS SDK to 17.6.1.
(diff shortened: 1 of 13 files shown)


ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -18,37 +18,53 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
         .notificationStatusChanged: "com.airship.flutter/event/notification_status_changed"
     ]
 
-    private static let streams: [AirshipProxyEventType: AirshipEventStream] = {
+
+    private static let eventSubject = PassthroughSubject<AirshipProxyEventType, Never>()
+    private static let eventTask: Task<Void, Never> =  Task {
```
rlepinski, line 23: Uses a task to send to a subject so we can have multiple listeners

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -18,37 +18,53 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
...
+        let pending = await AirshipProxyEventEmitter.shared.pendingEventAdded
+        for await event in pending {
+            eventSubject.send(event.type)
+        }
+    }
+
+    private let streams: [AirshipProxyEventType: AirshipEventStream] = {
```
rlepinski, line 30: Made these a local variable so we can listen from multiple sources

crow:
Static constants are lazily initialized on their first access. It looks like both eventTask and eventSubject are still uninitialized on their first use which causes pending events processing to not occur for incoming events like push received, message center displays, etc. Initializing eventTask and eventSubject to a discardable result on init (or on register...) seems to resolve this issue. Will push a fix when I have access.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Flutter Plugin Changelog
 
+## Version 7.2.0 - November 27, 2023
+Minor release that improves Feature Flag support on iOS.
+
+### Changes
+- Updated Airship iOS SDK to 17.6.1.
```
rlepinski, line 7: Updated Airship Android SDK to 17.5.0

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -18,37 +18,53 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
         .notificationStatusChanged: "com.airship.flutter/event/notification_status_changed"
     ]
 
-    private static let streams: [AirshipProxyEventType: AirshipEventStream] = {
+
+    private static let eventSubject = PassthroughSubject<AirshipProxyEventType, Never>()
+    private static let eventTask: Task<Void, Never> =  Task {
+        let pending = await AirshipProxyEventEmitter.shared.pendingEventAdded
+        for await event in pending {
+            eventSubject.send(event.type)
```
rlepinski, line 26: We should be able to remove this now that `AirshipProxyEventEmitter` exposes a subject for pending

rlepinski: You need to remove the publisher as well here and then have the streams use AIrshipProxyEventEmitter.shared.eventSubject (i should of exposed a publisher but oh well, its internal).

crow:
waiting on the latest airship framework proxy release (5.0.2) before re-running tests

crow (approved):

jyaganeh (approved):

PR #195, opened 2024-02-15
[MOBILE-4148] Fix gradle 8 build issue

Ulrico972:
### What do these changes do?
Add a namespace to our flutter module

To fix a build issue due to the use of gradle 8.

Reproduced on the flutter sample.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Flutter Plugin Changelog
 
+## Version 7.3.1 - February 16, 2024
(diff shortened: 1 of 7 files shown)


jyaganeh (approved):
We can also remove the `package` attribute from the manifest on [this line](https://github.com/urbanairship/airship-flutter/blob/main/android/src/main/AndroidManifest.xml#L2) now that we're setting it in gradle

rlepinski (approved):

PR #197, opened 2024-02-22
setup gcloud to get gsutil installed

Ulrico972:
Fixes releaseDoc job in release workflow by setting up gcloud to get gsutil installed.


jyaganeh (approved):
🆒

PR #205, opened 2024-05-30
Add compileOptions to fix new flutter release warning

tun43p:
### What do these changes do?
I've updated the sourceCompatibility and targetCompatibility for Java in the build.gradle file.

These changes are necessary since Flutter has updated its version and we have a warning in the console. Without this change, we can't update the configuration.

```txt
You are applying Flutter's app_plugin_loader Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/go/flutter-gradle-plugin-apply

You are applying Flutter's main Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/go/flutter-gradle-plugin-apply
```

I used my package instead of the pub.dev one.

```txt
✓ Built build/app/outputs/flutter-apk/app-staging-debug.apk
```

I'm open to discussion if you need more information.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -43,6 +43,10 @@ android {
     lintOptions {
         disable 'InvalidPackage'
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
+    }
 
     namespace = 'com.airship.airship'
 }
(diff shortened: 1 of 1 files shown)


rlepinski:
This looks good to us, we probably want to set a kotlin version as well. Thank you for the PR, we will try to get it merged this coming week

crow:
Going to merge this and add a couple remaining bits before we do a release. Thanks again for the PR

PR #210, opened 2024-07-25
Embedded views - Basic wiring with simple views

crow:
### What do these changes do?
Add basic initial support for embedded views.

To support embedded views in flutter.

Manual testing.

iOS:
https://github.com/user-attachments/assets/eff35227-a7ed-47a1-b270-b33aa5e31a68

Android:
https://github.com/user-attachments/assets/f9611935-bd9d-4306-a436-79b058d7a67f


android/src/main/kotlin/com/airship/flutter/EmbeddedViewFactory.kt
```diff
@@ -0,0 +1,21 @@
+package com.airship.flutter
+
+import android.content.Context
+import io.flutter.plugin.common.BinaryMessenger
+import io.flutter.plugin.common.MethodChannel
+import io.flutter.plugin.common.StandardMessageCodec
+import io.flutter.plugin.platform.PlatformView
+import io.flutter.plugin.platform.PlatformViewFactory
+
+class EmbeddedViewFactory(private val binaryMessenger: BinaryMessenger) : PlatformViewFactory(StandardMessageCodec.INSTANCE) {
```
jyaganeh, line 10: We should be able to include this class in the other file, like we do in Swift. If flutter doesn't like that for some reason, there's probably a compat annotation that will make it work. lmk if that's the case, I'm happy to help find the right one(s)

lib/src/airship_flutter.dart
```diff
@@ -1,38 +1,51 @@
...
   static final analytics = AirshipAnalytics(_module);
+
   /// The [AirshipActions] instance.
   static final actions = AirshipActions(_module);
+
   /// The [AirshipFeatureFlagManager] instance.
   static final featureFlagManager = AirshipFeatureFlagManager(_module);
 
+  /// The [AirshipAutomation] instance.
+  static final automation = AirshipAutomation(_module);
```
rlepinski, line 44: You should just use the AirshipInApp here

lib/src/airship_automation.dart
```diff
@@ -0,0 +1,69 @@
...
+  final Map<String, StreamController<bool>> _isEmbeddedAvailableControllers =
+      {};
+  List<EmbeddedInfo> _embeddedInfos = [];
+  late StreamSubscription<EmbeddedInfoUpdatedEvent> _subscription;
+
+  AirshipAutomation(this._module) {
+    _subscription = onEmbeddedInfoUpdated.listen(_updateEmbeddedIds);
+  }
+
+  bool isEmbeddedAvailable({required String embeddedId}) =>
```
rlepinski, line 42: If this is async does it cause issues?

android/src/main/kotlin/com/airship/flutter/InboxMessageViewFactory.kt
```diff
@@ -1,17 +0,0 @@
-package com.airship.flutter
-
-import android.content.Context
-import io.flutter.plugin.common.BinaryMessenger
-import io.flutter.plugin.common.MethodChannel
-import io.flutter.plugin.common.StandardMessageCodec
-import io.flutter.plugin.platform.PlatformView
-import io.flutter.plugin.platform.PlatformViewFactory
-
-class InboxMessageViewFactory(private val binaryMessenger: BinaryMessenger) : PlatformViewFactory(StandardMessageCodec.INSTANCE) {
```
rlepinski, line 10: Why was this removed?

rlepinski:
Android needs the same multicasting changes you made on iOS

jyaganeh (approved):

android/src/main/kotlin/com/airship/flutter/InboxMessageViewFactory.kt
crow: I have reformat on save, just spacing

crow: I just consolidated these the same way the embedded view was consolidated

rlepinski (approved):

PR #212, opened 2024-08-15
Fix message center message retrieval when launched from mc push

crow:
### What do these changes do?
Attempts a refresh when launching an MC message from a push. Similar to what we do in the react-native plugin.

To resolve some preventable message open failures.

Tested with and without the change launching the message center from an MC-containing push.

### DIFF
--- a/ios/Classes/AirshipInboxMessageView.swift
+++ b/ios/Classes/AirshipInboxMessageView.swift
@@ -86,7 +86,7 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, NativeBridgeDeleg
 
         if message == nil {
             /// Attempt a refresh is the message isn't available - as can happen when launched from a push
-            let success = try? await inbox.refreshMessages(timeout: 5)
+            let success = try? await inbox.refreshMessages(timeout: 100)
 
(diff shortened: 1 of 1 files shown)


ilitvinenko (approved):

ios/Classes/AirshipInboxMessageView.swift
```diff
@@ -79,8 +79,24 @@ class AirshipInboxMessageView : NSObject, FlutterPlatformView, NativeBridgeDeleg
...
         }
-        
+
         let inbox = MessageCenter.shared.inbox
+
+        let message = await inbox.message(forID: messageId)
+
+        if message == nil {
+            /// Attempt a refresh is the message isn't available - as can happen when launched from a push
+            let success = try? await inbox.refreshMessages(timeout: 5)
```
rlepinski, line 89: Should we just wait forever if it fails ? What would the handling look like on the flutter side for this?

crow: yeah probably - the async call just never receives a result, which gives the user the chance on the flutter side to continue showing a loading dialog

PR #218, opened 2024-09-13
prepare patch release 7.8.2

Ulrico972:


ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -452,7 +452,8 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
             return try AirshipJSON.wrap(flag).unWrap()
 
         case "featureFlagManager#trackInteraction":
-            guard let arg = try? call.requireAnyArg() as? String else {
+            guard let argument = try? call.requireAnyArg(),
```
rlepinski, line 455: we have a requireStringArg, maybe that would clena things up?

Ulrico972: I'll try that

ios/Classes/SwiftAirshipPlugin.swift
```diff
@@ -452,10 +452,7 @@ public class SwiftAirshipPlugin: NSObject, FlutterPlugin {
             return try AirshipJSON.wrap(flag).unWrap()
 
         case "featureFlagManager#trackInteraction":
-            guard let arg = try? call.requireAnyArg() as? String else {
-                throw AirshipErrors.error("Call requires a String argument")
-            }
-
+            let arg = try call.requireStringArg()
```
Ulrico972, line 455: I also removed the optional try, I think that's the reason why it was failing before

rlepinski (approved):

PR #220, opened 2024-10-19
Add LiveActivities and LiveUpdate support, push and message center updates

crow:
### What do these changes do?
[v] Update to proxy 10
[v] Show message center methods
[v] Fallback for enable user notifications
[v] Add notification permission status in the push notification status object
[v] Add live activity support
[v] Add live update support
[v] Notification service extension in example

To be in parity with react-native

Added UI to each example and tested against console to ensure live activities/updates were started and stopped as expected:

image

https://github.com/user-attachments/assets/b2531400-b0ca-411a-846f-2f6d1027e574


android/src/main/kotlin/com/airship/flutter/AirshipPluginExtender.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import android.content.Context
+import com.urbanairship.UAirship
+
+/**
+ * Extender that will be called during takeOff to customize the airship instance.
+ * Register the extender fully qualified class name in the manifest under the key
+ * `com.urbanairship.flutter.AIRSHIP_EXTENDER`.
+ */
+@Deprecated("Use com.urbanairship.android.framework.proxy.AirshipPluginExtender instead and register it under the manifest key `com.urbanairship.plugin.extender`")
```
rlepinski, line 13: We dont need this, only React had this before

rlepinski: delete the file

example/ios/ExampleWidgets/ExampleWidgets.swift
```diff
@@ -0,0 +1,79 @@
+/* Copyright Airship and Contributors */
+
+import WidgetKit
+import SwiftUI
+
+struct Provider: AppIntentTimelineProvider {
```
rlepinski, line 6: This is for other widgets i think

rlepinski (approved):

jyaganeh (approved):

PR #237, opened 2024-12-08
Add release preparation action and scripts

crow:
Release prep workflow and scripts


crow:
merging this to test it out

PR #245, opened 2025-01-06
MOBILE-4771: Research Flutter SPM support

ilitvinenko:
### What do these changes do?
- removes the obj-c version of plugin wrapper 
- adds support for SPM
- adds different imports for cocoapods and SPM use case
- added privacy manifest

https://urbanairship.atlassian.net/browse/MOBILE-4771

run flutter with SPM disabled (i.e. the old cocoapods way), make sure the app runs and archives with no problem
run flutter with SPM, make sure the app runs and archives with no problem

the proxy pr must be merged before https://github.com/urbanairship/airship-mobile-framework-proxy/pull/75 as these changes require more classes exposed from the proxy via SPM


ios/airship_flutter/Package.swift
```diff
@@ -0,0 +1,26 @@
...
+import PackageDescription
+
+let package = Package(
+    name: "airship_flutter",
+    platforms: [.macOS(.v10_15), .iOS(.v14), .tvOS(.v14), .visionOS(.v1)],
+    products: [
+        .library( name: "airship-flutter", targets: ["airship_flutter"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/urbanairship/airship-mobile-framework-proxy.git", from: "11.2.2")
```
ilitvinenko, line 13: need to set the correct new version of the proxy once https://github.com/urbanairship/airship-mobile-framework-proxy/pull/75 is merged and released

crow:
Looks good - we might want to get the proxy released and maybe generate/set up an spm_example variant? I just attempted this with the local version of the proxy but I'm not able to get it running.

example/ios/Runner/AppDelegate.swift
```diff
@@ -2,7 +2,12 @@
 
 import UIKit
 import Flutter
```
crow, line 4: does anything in this example actually need to change?

rlepinski (approved):

PR #260, opened 2025-07-11
[EUCUST-125] Remove duplicate resolve

Apekka:
### What do these changes do?
Remove the resolve function call in the feature flag part of the plugin.

Following an EU escalation, where the customer had crashes if the app was in airplane mode, it couldn't get the feature flags, and also couldn't catch the exception to avoid the crash.
cf logs : 
image
This happened cause we were calling twice result.error (in AirshipPlugin.kt and in Extensions.kt)

Running in the sample app, with airplane mode and trying to catch the exception of a feature flag, it works properly now we only have logs and the app can still function.

```
E/AirshipPlugin(30247): Error processing featureFlag#flag
E/AirshipPlugin(30247): com.urbanairship.featureflag.FeatureFlagException$FailedToFetch: Failed to fetch feature flag: 'rad_flag'! Stale data is not allowed.
```

### DIFF
--- a/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
+++ b/android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
@@ -304,11 +304,8 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             // Feature Flag
             "featureFlagManager#flag" -> result.resolve(scope, call) {
                 val args = call.jsonArgs().requireMap()
-                val flagName = args.get("flagName")?.requireString()
+                val flagName = requireNotNull(args.get("flagName")?.requireString())
                 val useResultCache = args.get("useResultCache")?.getBoolean(false) ?: false
-                if (flagName == null) {
-                    throw Exception("Missing flagName")
-                }
                 proxy.featureFlagManager.flag(flagName, useResultCache)
             }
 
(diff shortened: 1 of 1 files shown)


ilitvinenko:
are we sure we don't won't to report an error on exception? looks a little suspicious :)

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -362,7 +362,8 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                     try {
```
rlepinski, line 362: we should not have a try here, its already handled by the resolve method extension

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -251,145 +246,92 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
-                try {
-                    if (flagName == null) {
-                        throw Exception("Missing flagName")
-                    }
-
-                    proxy.featureFlagManager.flag(flagName, useResultCache)
-                } catch (e: Exception) {
-                    result.error("ERROR", "Failed to get flag", e.localizedMessage)
+                if (flagName == null) {
+                    throw Exception("Missing flagName")
```
rlepinski, line 310: Nit: requireNotNull instead

rlepinski (approved):

ilitvinenko (approved):

PR #268, opened 2025-08-21
[MOBILE-5230] Add check for flutter state before emitting events

rlepinski:
Adding tracking for if the activity and engine is attached before emitting events. UI events will look for if the activity is attached and the rest will look for just the engine being attached.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Flutter Plugin Changelog
 
+## Version 10.8.0 - August 28, 2025
+Minor release that updates the Android SDK to 19.11.0 and the iOS SDK to 19.8.3.
+
+### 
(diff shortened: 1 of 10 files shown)


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -326,10 +345,6 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
                 proxy.featureFlagManager.resultCache.cache(flag, milliseconds)
             }
 
-            "featureFlagManager#resultCacheRemoveFlag" -> result.resolve(scope, call) {
```
rlepinski, line 329: this was handled multiple times. other veersion is on line 336

PR #269, opened 2025-09-09
Add setForegroundPresentationOptionsCallback WIP

khmMouna:


android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -70,8 +77,31 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
             EventType.FOREGROUND_PUSH_RECEIVED to "com.airship.flutter/event/push_received",
             EventType.BACKGROUND_PUSH_RECEIVED to "com.airship.flutter/event/background_push_received",
             EventType.NOTIFICATION_STATUS_CHANGED to "com.airship.flutter/event/notification_status_changed",
-            EventType.PENDING_EMBEDDED_UPDATED to "com.airship.flutter/event/pending_embedded_updated"
+            EventType.PENDING_EMBEDDED_UPDATED to "com.airship.flutter/event/pending_embedded_updated",
+            EventType.OVERRIDE_FOREGROUND_PRESENTATION to "com.airship.flutter/event/override_presentation_options"
         )
+
+        @Volatile
+        var isOverrideForegroundDisplayEnabled: Boolean = false
```
crow: putting this in the companion object makes it static state that survives plugin detach/reattach. RN keeps it as an instance field — should move it down next to `foregroundDisplayRequestMap`

android/src/main/kotlin/com/airship/flutter/AirshipPlugin.kt
```diff
@@ -70,8 +77,31 @@ class AirshipPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
+    }
+
+    private val foregroundDisplayRequestMap =
+        ConcurrentHashMap<String, CompletableDeferred<Boolean>>()
+    private val foregroundDisplayPredicate = object : SuspendingPredicate<Map<String, Any>> {
+        override suspend fun apply(value: Map<String, Any>): Boolean {
+            val deferred = CompletableDeferred<Boolean>()
+            Log.d("AirshipPlugin", "apply() called, isOverrideForegroundDisplayEnabled=$isOverrideForegroundDisplayEnabled")
```
crow: debug log cruft we probably don't want anymore, this fires on every forground push
```suggestion
            val deferred = CompletableDeferred<Boolean>()
```

PR #295, opened 2026-04-17
Full SPM support for iOS plugin

crow:
### What do these changes do?
Adds SPM support alongside the existing CocoaPods setup. Apps on Flutter 3.24+ with SPM enabled resolve the plugin via `Package.swift`; everyone else keeps using the podspec. Bumps the plugin's Flutter floor to 3.24 and pins the target to Swift 5 language mode so we don't get dragged into a Swift 6 concurrency migration yet.

Flutter's moving toward SPM as the default and apps that flip the flag were hitting resolution issues. MOBILE-5552.

Built the example app via SPM end to end (`flutter config --enable-swift-package-manager` then `flutter build ios`) and separately ran `pod lib lint` to confirm the CocoaPods path still works.

Follow-up for Flutter 3.41's FlutterFramework dep and Swift 6 migration tracked in MOBILE-5589. Also untracked `Package.resolved` to match ios-library and the framework proxy.


ios/.gitignore
```diff
@@ -8,6 +8,7 @@
 profile
 airship_flutter/.swiftpm/
 airship_flutter/.build/
+airship_flutter/Package.resolved
```
crow, line 11: sometimes people like to do this sometimes they don't

Ulrico972 (approved):

khmMouna (approved):

README.md
```diff
@@ -64,11 +64,13 @@ For a more detailed setup guide, please see the full [Getting Started Documentat
 
 | Airship Flutter Version | Airship SDK Version | Flutter Version | Support Status |
 | :---------------------- | :------------------ | :-------------- | :------------- |
-| **11.x**                | 20.x                | 3.0.2+          | **Active**     |
+| **11.x**                | 20.x                | 3.24.0+         | **Active**     |
```
rlepinski: we need a new 12.x release to change flutter version

PR #299, opened 2026-05-08
Extract docs publishing into a reusable shared workflow

jyaganeh:
## Summary

- Extracts `release-docs` and `deploy-docs` from `release.yml` into a new reusable workflow `.github/workflows/publish-docs.yml`
- The shared workflow can be triggered automatically via `workflow_call` from the release pipeline, or manually via `workflow_dispatch` for a specific tagged version
- Consolidates Slack notifications into a single `notify-release` job that covers success, failure, and cancellation across all pipeline jobs

## Test plan

- [ ] Verify a tag push runs the full release pipeline and the `publish-docs` shared workflow is called correctly
- [ ] Verify a manual `workflow_dispatch` run of Publish Docs with a valid version tag checks out the right code and builds docs
- [ ] Verify the Slack notification fires correctly on success and failure
- [ ] Verify a `deploy-docs` failure (GitHub Pages not yet enabled) does not cause the Slack notification to report failure

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,73 @@
+name: Publish Docs
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+    secrets:
+      GCP_SA_KEY:
+        required: true
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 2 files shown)


crow:
yes this is what we need

crow (approved):

PR #322, opened 2026-09-18
Release 16.0.0

Ulrico972:
## Summary
- Bumps the native Android and iOS SDKs to 21.0.2 via framework proxy 16.0.1
- **Breaking:** drops CocoaPods support on iOS in favor of Swift Package Manager only (`airship_flutter.podspec` removed), matching the native SDK's own CocoaPods removal
- **Breaking:** raises Android's `minSdkVersion` to 26, matching the SDK 21 floor
- Adds `featureFlagManager.status()`, `.waitRefresh()`, and a `.statusUpdates` event stream
- Fixes the example app's own direct `ios-library` SPM pin (was stuck at 20.0.2, conflicting with the new floor)

See `CHANGELOG.md` and `MIGRATION.md` for full details and upgrade steps.

## Test plan
- [x] `flutter analyze` / `dart analyze --fatal-infos` clean
- [x] `flutter test` — all 29 tests pass, including new feature flag status/waitRefresh coverage
- [x] `flutter build ios --debug --no-codesign` (device target) — builds the full example `Runner.app` against live proxy 16.0.0 + ios-library 21.0.1
- [x] Android plugin module (`:airship_flutter:compileDebugKotlin`) compiles clean against proxy 16.0.0
- [ ] Manual smoke test of feature flag status/waitRefresh on a device/simulator

🤖 Generated with [Claude Code](https://claude.com/claude-code)


oristanovic:
Reviewed this branch and pushed fixes in `fa720e9`..`2e606f7`.

**Consumer facing**
- `pubspec` declared `flutter: ">=3.24.0"` while the plugin is SPM-only. Below **3.44.0** (first stable with SPM enabled by default) `flutter pub get` fails on *every* platform, Android included — this branch already hit it, hence `--enable-swift-package-manager` in the docs and Android CI jobs. Raised to 3.44.0 / Dart 3.10.0, documented in CHANGELOG + MIGRATION.
- `AirshipEmbeddedView.windowHeight()` returned nil whenever no window reported as key, leaving a self-sizing embedded view with no height proposal — the bug 12.5.0 fixed. It only ever needed a *screen* height and a scene knows its own screen, so the window lookup is gone; `@MainActor` restored.
- `featureFlagManager.status()` folded unknown values into `outOfDate`, hiding payload mismatches. Now throws, documented and tested.
- `waitRefresh()` sent `{"maxTimeMillis": null}` — the same null-in-a-map shape that broke takeOff on iOS 27. Now omits the key.

**Example app and CI**
- The example wasn't UIScene-migrated, so built with Xcode 27 it was terminated at scene creation on iOS 27 before any Dart ran. Added `SceneDelegate` + scene manifest.
- `ci-ios` only *built* the example, which is why that went unnoticed. `scripts/smoke_ios.sh` now boots a simulator, launches the app, and fails if the process died **or** takeOff errored — the second check matters, since EUCUST-160 left the app running and rendering.
- Regenerated the tracked `FlutterGeneratedPluginSwiftPackage/Package.swift`, stale for 3.44+ (`airship_flutter` → `airship-flutter`, plus `FlutterFramework`). This was the cause of `Missing package product` errors in Xcode.
- Declared `FlutterFramework` in the plugin's `Package.swift`, removing the warning printed on every consumer build. Verified it exists in 3.44.0 too, so the floor holds.
- Committed Flutter's auto-applied `android.builtInKotlin=false` / `android.newDsl=false`, and dropped the example's dead `AirshipKit` fallback.

**Verified**: `flutter analyze` + 31 tests; clean-DerivedData `xcodebuild` → BUILD SUCCEEDED with no FlutterFramework warning; iOS 27 simulator and a physical iOS 27 device both reach `TakeOff completed` (SDK 21.0.2); Android emulator builds, launches and runs clean. The smoke test was checked in **both** directions — removing the scene manifest and forcing takeOff to throw each fail it. Also confirmed this branch fixes EUCUST-160: on a clean Xcode 27 build, 12.6.0 fails takeOff with `Invalid JSON <null>` and this branch does not.

**Notes**: `ios/Flutter/ephemeral/` is tracked here although standard Flutter projects gitignore it, so it must be regenerated on every Flutter bump — that is what broke Xcode above. A pre-existing Swift 6 warning (`AirshipPlugin` conformance crossing main-actor isolation) is left alone. If you hit `Missing package product` after checkout, it's stale Xcode state: `rm -rf ~/Library/Developer/Xcode/DerivedData/Runner-*`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rlepinski (approved):

